PR #3, opened 2022-02-17
Upgrade from sixtyfps to slint UI framework

secana:
The UI framework used in the KDE example was renamed from "sixtyfps" to "slint". This PR switches from the old name to the new name, as the Crates under the old name won't receive any update in the future.


secana:
As we do not have an automated test for the examples yet, I've build and tested the change locally and it works.

unglaublicherdude (approved):

PR #4, opened 2022-02-18
switch from draw.io to mermaid

secana:
Github support Mermaid.js (https://mermaid-js.github.io/mermaid/#/) since a few days. It allows the generation of graphs directly in markdown instead the current approach where a draw.io project has to be exported to a svg.

This PR removes all draw.io diagrams and replaces them with inline mermaid markdown.


unglaublicherdude (approved):

unglaublicherdude:
I like it!

PR #12, opened 2022-03-11
Add cancellation for typescript sdk

secana:
This PR adds a cancellation token to the TypeScript SDK. With that token, it is possible to cancel a request to the VaaS API after a given time. The PR is not a breaking change, as a default value is set for all methods.

### DIFF
--- a/typescript/src/vaas.ts
+++ b/typescript/src/vaas.ts
@@ -1,15 +1,16 @@
 import WebSocket from "isomorphic-ws";
 import * as sha256 from "fast-sha256";
-import { Kind, Message } from "./messages/message";
-import { AuthenticationResponse } from "./messages/authentication_response";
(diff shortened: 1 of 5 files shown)


typescript/src/vaas.ts
```diff
@@ -34,64 +37,64 @@ export default class Vaas {
...
             }
-            const guid = uuidv4()
 
+            const guid = uuidv4()
             this.callbacks.set(guid, async (verdictResponse: VerdictResponse) => {
+                const timer = setTimeout(() => {
+                    reject("Cancelled");
+                }, ct.timeout());
+
                 if (typeof sample === "string") {
```
unglaublicherdude, line 82: that diff was not easy to read . :D

unglaublicherdude (approved):

PR #19, opened 2022-05-17
Dotnetsdk

sfotiadis:


doxthree:
Do you also plan to publish this sdk to nuget.org?

dotnet/Vaas/Vaas.sln.DotSettings.user
```diff
@@ -0,0 +1,12 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
pstadermann, line 1: This file should not be committed. You can use https://github.com/github/gitignore/blob/main/VisualStudio.gitignore as .gitignore

dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Text.Json.Serialization;
+using System.Threading;
+using System.Threading.Tasks;
+using Vaas.Messages;
+using Websocket.Client;
+using Websocket.Client.Exceptions;
+
+namespace Vaas
+{
+    public class Vaas
```
pstadermann, line 18: Add

/// <remarks>not thread-safe</remarks>

or make thread-safe. This would require changes (all properties are not thread-safe). Contact me for more infos.

dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -0,0 +1,181 @@
...
+                        break;
+                        
+                    case "VerdictResponse":
+                        var options = new JsonSerializerOptions {Converters = {new JsonStringEnumConverter()}};
+                        var verdictResponse = JsonSerializer.Deserialize<VerdictResponse>(msg.Text,options);
+                        VerdictResponsesDict.Add(verdictResponse?.Guid ?? throw new InvalidOperationException(), verdictResponse);
+                        break;
+                }
+            });
+            Client.Start().GetAwaiter().GetResult();
```
pstadermann, line 63: Client.Start().Wait()

dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -0,0 +1,181 @@
...
+            if (!Client.IsStarted)
+            {
+                throw new WebsocketException("Could not start client");
+            }
+            
+            Authenticate();
+            Task.Run(SendPing);
```
pstadermann, line 70: Remove. Websocket have their own keepalive mechanism (ping, pong, frames). The library takes care of that.

dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        private void Authenticate()
+        {
+            var authenticationRequest = new AuthenticationRequest(Token);
+            var jsonString = JsonSerializer.Serialize(authenticationRequest);
+            Client?.Send(jsonString);
+            for (var i = 0; i < 10; i++)
+            {
+                if (Authenticated)
```
pstadermann, line 80: Updating a bool from another thread may not become visible (if the variable gets optimized into a register). This was the case in the past with x86 release builds. This works for x64 builds, but is not guaranteed by the memory model.

tldr: This might break in the future.

dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            var sha256 = Sha256CheckSum(path);
+            var verdictResponse = await ForRequestAsync(new AnalysisRequest(sha256, SessionId ?? throw new InvalidOperationException()));
+            if (verdictResponse.Verdict != Verdict.Unknown) return verdictResponse.Verdict;
+            var url = verdictResponse.Url;
+            if (url is null) throw new ArgumentNullException(nameof(url));
+            
+            var token = verdictResponse.UploadToken;
+            var data = await File.ReadAllBytesAsync(path);
+            using (var client = new WebClient())
```
pstadermann, line 110: WebClient is obsolete. The modern version is HttpClient. Make it a field, otherwise it opens a new socket for every request.

dotnet/Vaas/src/Vaas/Messages/AuthenticationResponse.cs
```diff
@@ -0,0 +1,19 @@
...
+    public class AuthenticationResponse
+    {
+        [JsonPropertyName("kind")] 
+        public string Kind = "AuthResponse";
+        
+        [JsonPropertyName("success")] 
+        public bool Success { get; init; }
+
+        [JsonPropertyName("session_id")] 
+        public string SessionId { get; init; } = null!;
```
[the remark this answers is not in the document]
secana: That strange syntax come from the .NET deserializer implementation. The "*Response" types are never instantiated in code but only from the deserializer.

The value must not be null and has to throw an exception if a null value gets deserialized. On the other hand, it needs to be "init", such that that the deserializer can write to it, but it is immutable after. That's why "set" or only "get" won't work.

Setting it to the empty string won't work, as a missing field would be deserialized to an empty string, instead of an exception.

Not happy with that code,  but the only why Anni and me came up with a way to encode the requirements correctly in code. Not the strongest type system here...

I'm open to any more elegant solutions :)

dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -0,0 +1,181 @@
...
+        
+        private async Task<VerdictResponse> ForRequestAsync(AnalysisRequest analysisRequest)
+        {
+            var jsonString = JsonSerializer.Serialize(analysisRequest);
+            await Task.Run(()=>Client?.Send(jsonString));
+
+            return await WaitForResponseAsync(analysisRequest.Guid);
+        }
+
+        private void SendPing()
```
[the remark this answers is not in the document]
secana: Does the server send the ping now first? Test with the slow responding server showed that the connection is closed before a response get send. Ping keeps the connection open. If we guarantee, that the response is always faster than the timeout, we can remove it.

dotnet/Vaas/src/Vaas/Vaas.cs
secana: That's a good call. General question: Does it need to be thread-safe? Only SDK I'm sure that is thread-safe is currently the Rust one. Either we make the C# version thread-safe or we expect the user to create a new VaaS object per thread. Any case, needs to be documented.

dotnet/Vaas/src/Vaas/Vaas.cs
secana: Lib we use here unfortunatly not. It only reacts on a ping it gets but does not send any by itself. IMHO we implemented it wrong on the server side :(

dotnet/Vaas/src/Vaas/Vaas.cs
secana: As far as I know, it's better to use "GetAwaiter" as "Wait" gives you an AggregateException in every error case, where "GetAwaiter" gives you the underlying exception. But that's more of taste/nitpick

dotnet/Vaas/src/Vaas/Messages/AuthenticationResponse.cs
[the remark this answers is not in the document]
pstadermann: Auf diese Weise hast du leider "worst-of-both-worlds": Zwar kompiliert es, aber er schmeißt keine Exception, wenn die Property nicht im JSON existiert und der Wert kann nach der Deserialisierung null sein, obwohl der Typ nicht nullable ist. S. diesen Test. Weil ich mir auch nicht sicher war, habe ich das gerade getestet.

```C#
public class Model
{
    public string A { get; set; } = null!;
}

public class SerializerTest
{
    [Fact]
    public void ThrowsException()
    {
        string json = "{}";
        var model = JsonSerializer.Deserialize<Model>(json);
        Assert.Equal(null, model.A);
    }
}
```

Bei Json.net konnte man [Required] setzen. System.Text.Json hat noch keine Implementierung, aber eventuell hilft das https://github.com/dahomey-technologies/Dahomey.Json . Die Lib stellt Attribute zu Verfügung. Klappt auch.

dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -0,0 +1,181 @@
...
+        public async Task<List<Verdict>> ForFileListAsync(IEnumerable<string> fileList)
+        {
+            return (await Task.WhenAll(fileList.Select(ForFileAsync))).ToList();
+        }
+
+        
+        private async Task<VerdictResponse> ForRequestAsync(AnalysisRequest analysisRequest)
+        {
+            var jsonString = JsonSerializer.Serialize(analysisRequest);
+            await Task.Run(()=>Client?.Send(jsonString));
```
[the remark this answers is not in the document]
pstadermann: Send only enqueues and does not block.

dotnet/Vaas/src/Vaas/Vaas.cs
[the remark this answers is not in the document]
pstadermann: Both Client and Server have a keep-alive. The configured 30s in was too high (connection is disconnected after 20s). I set it to 20s.

dotnet/Vaas/src/Vaas/Vaas.cs
pstadermann: Thread-safe now

dotnet/Vaas/src/Vaas/Vaas.cs
pstadermann: Converted to async.

dotnet/Vaas/src/Vaas/Vaas.cs
pstadermann: Another possible reason could have been a Thread.Sleep(40s), which prevents all async code from running in the same thread. We encountered this with Python.

Added a test, works.

dotnet/Vaas/src/Vaas/Vaas.cs
pstadermann: Switched to Task.

dotnet/Vaas/src/Vaas/Messages/AuthenticationResponse.cs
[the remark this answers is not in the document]
pstadermann: It's a bit better now. For the future: Use proper model validation when System.Text.Json supports it.

pstadermann (approved):

PR #33, opened 2022-06-01
let devcontainer work correct with lombok (getter/setter)

unglaublicherdude:


unglaublicherdude:
Already implemented in [PR 32](https://github.com/GDATASoftwareAG/vaas/pull/34)

PR #77, opened 2022-08-17
Add custom guid

GdataGithubBot:


typescript/tests/test.ts
```diff
@@ -41,6 +41,10 @@ function createVaas() {
 
 const defaultTimeout: number = 10_000;
 
+function test(foobar: string = "nix") {
```
pstadermann, line 44: Remove

pstadermann (changes requested):
Looks good. Only 1 dead code.

unglaublicherdude:
maybe we should also move all the forX tests to the new ones with custom guids. Will have a look at this.

PR #78, opened 2022-08-23
PHP Tests: outputs the tests uuids

unglaublicherdude:


unglaublicherdude:
The PHP SKD already provided the capability to set your own UUID (and provide it as parameter). Alls this does is output it while testing, so we can check tracing for these uuids.

unglaublicherdude:
Put a bit more work into it. Because the Functions did not return the uuid so the customer had no chance to compare the it to its requests. Therefore I split up into to function one that returns the verdictResponse object and one that just calls that funtion and only returns the verdict as string.

unglaublicherdude:
I keept the old ones mostly for compatibility reasons.

doxthree (approved):

pstadermann (approved):

Issue #97, opened 2022-09-19
Composer expects PHP SDK in root folder of repository


secana:
The `composer.json` needs to be in the root of the SDK to work correctly. Currently Composer cannot include the PHP SDK as the file is in a sub-folder. Relates to #96

PR #101, opened 2022-09-26
Move composer-Files to make php package installable

dr-bock:
See if you want to adopt it that way or prefer another solution. But:

If you want the PHP people to be able to install/clone the SDK directly from the repo using the composer PackageManager, then the composer.json file must be in the root of the repository. Otherwise composer will not be able to find the package. There is unfortunately no way to tell composer to look in a subdirectory of a repository.

Additional Notes:
If you also want to make the package publicly available, you can publish it via https://packagist.org/. (Only the github-repo will be linked).


secana:
Fixes #97

Issue #93, opened 2022-09-13
Zero-byte file throws HTTP 403 "Content Length not set"


secana:
Reproduce:

Try to check a zero byte (empty) file with VaaS Rust SDK. Instead of the expected clean verdict, the following is returned:

```bash
ERROR gdata_cloud_scanner::error: Failed to scan files: Server answered with status code: `400 Bad Request` `{"message":"Bad request: Content-Length not set"}`
```

Not sure, if the error is Rust specific, but we should add a test-case for all SDKs, which test the empty file case.

secana:
I've added tests to all SDKs in the `no-content-branch`. The error has to be fixed on the server side and then the SDKs can be checked again.

secana:
I created a PR to the backed project, which fixes the empty file response and also moved the empty file from the `unknown` to the `clean` state in the database.

The server changes need to be merged into stating and then into production to take effect. The database change should take effect, as soon as the caches are cleared.

secana:
I've checked that `rust`, `python` and `dotnet` work, but the following languages have broken certificate checks and do not use the imported staging cert, so they are untested: `java`, `php`, `typescript`. We will have to check that if the server code is merged into master.

secana:
Code is merged into master :)

PR #109, opened 2022-10-12
Tests: No content bug

secana:
Add tests for #93 

The tests will fails, as the fix is currently only in `staging` on the server side. As soon, as they are merged into `production`, the test should succeed.


unglaublicherdude (approved):

typescript/tests/test.ts
```diff
@@ -1,27 +1,27 @@
-import { expect } from "chai";
```
pstadermann, line 1: Editorconfiguration is wrong. Prettier is not used. Nothing should be reformatted. Did you use the devcontainer?

pstadermann (changes requested):

typescript/tests/test.ts
secana: Fix the formatting merge conflict. I don't use the container. I use the default from whatever JetBrains uses. Is there any non-default formatting that the team committed to?

pstadermann: If you install prettier, it takes care of everything on save.
For IntelliJ you can download the plugin here (not tested): https://www.jetbrains.com/help/idea/prettier.html

doxthree (approved):

PR #133, opened 2022-11-09
Python/get verdict for url

unglaublicherdude:


secana:
In this approach the SDK downloads the content of the URL and than uploads that over that to the VaaS backend. IMHO the SDK should submit the URL to the VaaS backend, which does the download. Am I missing something here?

unglaublicherdude:
> In this approach the SDK downloads the content of the URL and than uploads that over that to the VaaS backend. IMHO the SDK should submit the URL to the VaaS backend, which does the download. Am I missing something here?

me no comprendo. The for_url function does put the url into a new message 'VerdictRequestForUrl' 
https://github.com/GDATASoftwareAG/vaas/pull/133/files#diff-4eebc16016247140146556a7236af9c3cbb46d7e70ccd93b6e32ebb2092f51ddR293

It is not downloaded there.

doxthree (approved):

PR #162, opened 2023-01-23
ruby send lists instead of single requests

AllieW99:


AllieW99:
Closing PR because this was not the needed feature

PR #165, opened 2023-02-08
Java/tenant

unglaublicherdude:


java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -63,61 +61,55 @@ public void disconnect() throws InterruptedException {
         }
     }
 
-    public CompletableFuture<VerdictResponse> forUrlAsync(URL url) throws Exception {
+    public CompletableFuture<VerdictResponse> forUrlAsync(URL url, HashMap<String, String> verdictRequestAttributes) throws Exception {
```
secana, line 64: General question: Do we expect a user of the SDK to always provide tenant info? I would assume no, in which case the API gets significantly more difficult to use.

Does it make sense to have new API function with the support of attributes instead of adding it to the existing functions? This would also make a breaking change a non-breaking change as no update on the users side is needed.

java/src/main/java/de/gdata/vaas/VaasConfig.java
```diff
@@ -33,4 +38,9 @@ public VaasConfig(URI tokenEndpoint, URI url) {
         this.url = url;
         this.PullDelayMs = 100;
     }
+
+    public VaasConfig(URI tokenEndpoint, URI url, Duration timeout) {
```
secana, line 42: Not sure about Java here, but if you set a defaultTimeout in line 28 and override it here anyways, way set it at all? I would not set the timeout in the constructor at all, or provide a second constructor with that argument.

secana (changes requested):

PR #168, opened 2023-02-15
Fix broken export

### LINKED ISSUE #167: ClientCredentialsGrantAuthenticator not exported correctly
I've tried to use the npm package with Deno for an example, but the `ClientCredentialsGrantAuthenticator` is not exported correctly. This breaks the example on the Readme.

```typescript
import { Vaas, Verdict, ClientCredentialsGrantAuthenticator } from "npm:gdata-vaas@4.0.2";
```

Results in:

```bash
error: Uncaught SyntaxError: The requested module 'npm:gdata-vaas@4.0.2' does not provide an export named 'ClientCredentialsGrantAuthenticator'
import { Vaas, Verdict, ClientCredentialsGrantAuthenticator } from "npm:gdata-vaas@4.0.2";
```

secana:
Fix #167 and clean up the deprecated code.


secana:
As I've removed the deprecated code, this is a breaking change which needs a major version bump.

typescript/package.json
```diff
@@ -39,14 +39,12 @@
...
   "dependencies": {
     "@types/uuid": "8.3.4",
-    "@types/ws": "8.5.3",
     "@ungap/url-search-params": "0.2.2",
     "axios": "1.2.1",
     "fast-sha256": "1.3.0",
     "isomorphic-ws": "5.0.0",
     "typescript-json-serializer": "5.1.0",
-    "uuid": "9.0.0",
-    "ws": "8.11.0"
```
pstadermann, line 49: ws should be a dependency if we are not bundling.

typescript/tsconfig.json
```diff
@@ -68,7 +68,7 @@
 
     /* Interop Constraints */
     // "isolatedModules": true,                          /* Ensure that each file can be safely transpiled without relying on other imports. */
-    // "allowSyntheticDefaultImports": true,             /* Allow 'import x from y' when a module doesn't have a default export. */
+    "allowSyntheticDefaultImports": true,             /* Allow 'import x from y' when a module doesn't have a default export. */
```
pstadermann, line 71: Why is this necessary?

typescript/src/ClientCredentialsGrantAuthenticator.ts
```diff
@@ -1,6 +1,6 @@
 import getTokenWithClientCredentialsGrant from "./getTokenWithClientCredentialsGrant";
 
-export default class ClientCredentialsGrantAuthenticator {
+export class ClientCredentialsGrantAuthenticator {
```
pstadermann, line 3: Leave as is.

typescript/src/Index.ts
```diff
@@ -1,6 +1,5 @@
 export * from "./Vaas";
-export * from "./CreateVaasWithClientCredentialsGrant";
 export * from "./Verdict";
 export * from "./CancellationToken";
 export * from "./VaasErrors";
-export { default as ClientCredentialsGrantAuthenticator } from "./ClientCredentialsGrantAuthenticator";
+export * from "./ClientCredentialsGrantAuthenticator";
```
pstadermann, line 5: This shouldn't require a change.

typescript/tests/createVaas.ts
```diff
@@ -11,17 +11,21 @@ function getFromEnvironment(key: string) {
...
+  const authenticator = new ClientCredentialsGrantAuthenticator(
     CLIENT_ID,
     CLIENT_SECRET,
     TOKEN_URL,
-    VAAS_URL
   );
+
+  let vaas = new Vaas();
+  let token = await authenticator.getToken()
+  await vaas.connect(token, VAAS_URL);
```
pstadermann, line 29: Change in behaviour. Isn't This is different to the way other libs do it.

typescript/tests/createVaas.ts
```diff
@@ -11,17 +11,21 @@ function getFromEnvironment(key: string) {
...
 
-  return CreateVaasWithClientCredentialsGrant(
+  const authenticator = new ClientCredentialsGrantAuthenticator(
     CLIENT_ID,
     CLIENT_SECRET,
     TOKEN_URL,
-    VAAS_URL
   );
+
+  let vaas = new Vaas();
```
pstadermann, line 27: const

typescript/tests/createVaas.ts
```diff
@@ -11,17 +11,21 @@ function getFromEnvironment(key: string) {
...
+  let token = await authenticator.getToken()
```
pstadermann, line 28: const

pstadermann (changes requested):

typescript/src/ClientCredentialsGrantAuthenticator.ts
unglaublicherdude: this is the major change isn't it?

> Leave as is.

typescript/tests/createVaas.ts
unglaublicherdude, line 28:
```suggestion
  const token = await authenticator.getToken()
```

unglaublicherdude: you can just make suggestions
```suggestion
  const token = await authenticator.getToken()
```

typescript/tests/createVaas.ts
unglaublicherdude:
```suggestion
  const vaas = new Vaas();
```

typescript/tests/createVaas.ts
unglaublicherdude: isn't this just exaclty what the old code in typescript/src/CreateVaasWithClientCredentialsGrant.ts did?

typescript/src/ClientCredentialsGrantAuthenticator.ts
unglaublicherdude, line 3: @secana so when having an 

`export default'`

you don't need the

`"allowSyntheticDefaultImports":` true `

flag right? Is there any reason to remove the default?

PR #199, opened 2023-03-24
Remove broadcast and use simple channels in go sdk

Henning1:


secana:
@Henning1 is the draft still in progress?

Henning1:
> @Henning1 is the draft still in progress?

Some requests have long response times (seconds), which needs further investigition. Otherwise the code is ready for use.

sfotiadis (approved):

PR #224, opened 2023-05-08
Change all staging URLs

borisahrens:


unglaublicherdude:
We decided not to care about typescript for this release. and open an issue to fix the error there.

PR #235, opened 2023-05-24
Java/build for java11

unglaublicherdude:


unglaublicherdude:
It is called legacy in the ci because it support ends in 4 month.

https://endoflife.date/java

PR #259, opened 2023-07-20
java/105-useshed-usecache-flags 

doxthree:
- Add VaasOptions with UseCache & UseShed Flags + default values
- Add functionality to add custom guid for verdict request + url
- Apply best practice for javadoc

### DIFF
--- a/java/src/main/java/de/gdata/vaas/messages/VerdictRequestForUrl.java
+++ b/java/src/main/java/de/gdata/vaas/messages/VerdictRequestForUrl.java
@@ -27,8 +27,8 @@ public class VerdictRequestForUrl extends MessageType {
     @SerializedName("verdict_request_attributes")
     VerdictRequestAttributes verdictRequestAttributes;
     @Getter
-    @SerializedName("use_shed")
-    transient boolean UseShed;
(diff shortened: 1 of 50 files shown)


java/src/main/java/de/gdata/vaas/messages/VerdictRequestForUrl.java
```diff
@@ -26,22 +27,36 @@ public class VerdictRequestForUrl extends MessageType {
     @SerializedName("verdict_request_attributes")
     VerdictRequestAttributes verdictRequestAttributes;
     @Getter
-    @Setter
+    @SerializedName("use_shed")
+    transient boolean UseShed;
```
unglaublicherdude, line 31: transient?

unglaublicherdude (approved):

java/src/main/java/de/gdata/vaas/messages/VerdictRequestForUrl.java
secana: I would prefer not to call it `use_shed` but something like `use_hash_lookup` or something that does not leak internal component names and describes more what it does instead.

PR #277, opened 2023-08-14
Improve error handling for Java SDK #276

lennartdohmann:


java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -174,13 +188,19 @@ public VaasVerdict forSha256(Sha256 sha256) throws ExecutionException, Interrupt
      *                                       is closed
      */
```
pstadermann, line 189: Add VaasClientException, VaasServerException

PR #308, opened 2023-10-11
Ruby password authentication

AllieW99:


lennartdohmann (changes requested):
The example should run as GitHub workflow

AllieW99:
> The example should run as GitHub workflow

Done. It's tested. It will work perfectly when the gem is released.

lennartdohmann (approved):

PR #309, opened 2023-10-11
add resource_owner_password_grant_authenticator

LStapf:


pstadermann (approved):

secana:
As we know have multiple authentication methods, I would prefer if all authentication logic is moved to one module. This would solve the very long module name `resource_owner_password_grant_authenticator`, too. E.g. add a `auth` module and move all authentication logic into that. 

This is a breaking change, but more clean than the current approach where some of the logic is in `vaas` and some in `resource_owner_password_grant_authenticator`. 

I like the approach with the `Authentication` trait. That should be implemented for the current auth. method as well, such that it is consistent.

rust/src/resource_owner_password_grant_authenticator.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub(crate) struct ResourceOwnerPasswordGrantAuthenticator {
+    client_id: String,
+    user_name: String,
+    password: String,
+    token_url: Url,
+}
+
+impl ResourceOwnerPasswordGrantAuthenticator {
+    pub fn new(client_id: &str, user_name: &str, password: &str, token_url: &Url) -> Self {
```
secana, line 14: token_url should have a default value that the customer does not need to set. I suggest to use the builder pattern here, to allow it to be an option.

If you need to own all the values anyway, why expect a reference?

secana (changes requested):

rust/src/resource_owner_password_grant_authenticator.rs
pstadermann: TODO:
* New auth module
* Move all implemtations of Authentication to auth
* Implement Authentication for client_credentials_grant
* Add auth to builder
* Remove parameters from connect
* Methods: Pass values instead of references if ownership is required

pstadermann: * Use builder for Authenticator options (https://crates.io/crates/derive_builder), for easy to use values

secana:
I assigned the PR to me to implement the changes I had i mind.

secana (approved):

secana (approved):

Issue #354, opened 2024-02-05
Implement scan methods for streams


lennartdohmann:
Add a method to each SDK that allows to check for streams, similar to [Java's forInputStream](https://github.com/GDATASoftwareAG/vaas/blob/e9d91cb9f71cf51a264ac9c60c2b54b347288fdd/java/src/main/java/de/gdata/vaas/Vaas.java#L433) method.

- [x] Typescript
- [x] Go
- [x] Dotnet
- [x] Python
- [x] Rust
- [x] Ruby
- [x] PHP

In addition, tests and examples should be added in parallel to the existing ones in the examples folder.

unglaublicherdude:
ruby still missing

unglaublicherdude:
ruby will not be done

PR #356, opened 2024-02-08
Feature/354 for stream

lennartdohmann:
Implement scan methods for streams #354

### DIFF
--- a/dotnet/Vaas/src/Vaas/Vaas.cs
+++ b/dotnet/Vaas/src/Vaas/Vaas.cs
@@ -183,6 +183,69 @@ public class Vaas : IDisposable, IVaas
         return new VaasVerdict(verdictResponse);
     }
 
+    public async Task<VaasVerdict> ForStreamAsync(
+        Stream stream,
+        long contentLength,
+        CancellationToken cancellationToken,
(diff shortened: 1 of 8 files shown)


dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -183,6 +183,68 @@ public async Task<VaasVerdict> ForUrlAsync(Uri uri, CancellationToken cancellati
...
+            );
+        }
+
+        var response = WaitForResponseAsync(verdictResponse.Guid);
+        await UploadStream(stream, verdictResponse.Url, verdictResponse.UploadToken, cancellationToken);
+        
+        return new VaasVerdict(await response);
+    }
+    
+    private async Task UploadStream(Stream stream, string url, string token, CancellationToken cancellationToken)
```
doxthree, line 223: I noticed that this function is very similar with this function here https://github.com/GDATASoftwareAG/vaas/blob/d975d583ee41bfa549c8591ab3a47741e21899a6/dotnet/Vaas/src/Vaas/Vaas.cs#L335
Any way to combinate these two since both are working with Streams?

unglaublicherdude (approved):

PR #358, opened 2024-02-08
Typescript/add for stream method

doxthree:


pstadermann:
I'd expect forStream and the tests to use
* https://nodejs.org/api/stream.html
* https://wanago.io/2019/03/11/node-js-typescript-5-writable-streams-pipes-and-the-process-streams/

pstadermann:
https://medium.com/@barrytwice/node-js-streams-the-ultimate-guide-99af054c60c8

typescript/src/Vaas.ts
```diff
@@ -215,6 +237,34 @@ export class Vaas {
...
+      const guid = uuidv4();
+      if (this.debug) console.debug("uuid", guid);
+      this.verdictPromises.set(guid, {
+        resolve: async (verdictResponse: VerdictResponse) => {
+          if (verdictResponse.verdict === Verdict.UNKNOWN) {
+            const contentLength = stream.readableLength;
+            await this.upload(verdictResponse, stream, contentLength);
+            this.verdictPromises.delete(guid);
+          }
+          resolve(new VaasVerdict(verdictResponse.sha256, verdictResponse.verdict));
```
lennartdohmann, line 255: Doesn't this mean that if Malicious or Clean comes back, the user gets a Verdict without the content having been seen by the server? This shouldn't happen in theory, but just in case, we throw a VaasServerException here in the other SDKs...

lennartdohmann (changes requested):

lennartdohmann (approved):

PR #362, opened 2024-02-16
remove php 7.4 and 8.0 as they are eol.

unglaublicherdude:
also removed an unused dependency


unglaublicherdude:
There were also some formatting rules applied. Please if you contribute to the php sdk use the devcontainers, in the php folder.

PR #367, opened 2024-02-21
adds for stream method

unglaublicherdude:


golang/vaas/pkg/vaas/vaas.go
```diff
@@ -355,6 +357,62 @@ func (v *vaas) ForUrl(ctx context.Context, url string) (msg.VaasVerdict, error)
 	}, nil
 }
 
+// ForUrl sends an analysis request for a file URL to the Vaas server and returns the verdict.
```
pstadermann, line 360: forStream sends ...

pstadermann (changes requested):

PR #377, opened 2024-02-29
Rust: implements the cache and hash-lookup flags

unglaublicherdude:
this is just a draft because when the forStream pr is merged the flags have to be implemented for that too


secana (approved):
looks good to me

Issue #390, opened 2024-03-11
Support Detections and file/mime type


pstadermann:
The verdict responses now return detection names and the file/mime type. Take a look at the [Java implementation](https://github.com/GDATASoftwareAG/vaas/pull/386/files)

SDK support:
* [x] Java
* [x] Dotnet
* [x] Rust
* [x] TypeScript
* [x] Go
* [x] PHP
* [x] Python
* [ ] #442

PR #393, opened 2024-03-12
PHP implementation of #390

lennartdohmann:
PHP implementation of #390


php/src/vaas/Message/VerdictResponse.php
```diff
@@ -9,4 +9,7 @@ class VerdictResponse extends BaseMessage
     public string $guid;
     public string $sha256;
     public ?string $upload_token;
+    public ?LibMagic $libMagic;
+
```
unglaublicherdude, line 13: remove the empty line

PR #399, opened 2024-03-13
fixes the casing for detections and mime-types

unglaublicherdude:


unglaublicherdude:
I also fixed allot of warnings from the pipelines. Misuse, wrong parameters, updates.

PR #401, opened 2024-03-14
proposal to make the forStream function more clear

unglaublicherdude:
dunno, this is just a proposal as we broke our heads about that today

### DIFF
--- a/typescript/src/Vaas.ts
+++ b/typescript/src/Vaas.ts
@@ -241,22 +241,39 @@ export class Vaas {
       const guid = uuidv4();
       if (this.debug) console.debug("uuid", guid);
       let contentLength = 0;
-      this.verdictPromises.set(guid, {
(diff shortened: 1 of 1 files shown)


lennartdohmann (approved):

typescript/src/Vaas.ts
```diff
@@ -241,22 +241,39 @@ export class Vaas {
...
+            throw new VaasServerError(
+              "No upload token set in response",
+            );
+          }
+          if (verdictResponse.url === undefined || verdictResponse.url == "") {
+            throw new VaasServerError(
+              "No upload url set in response",
+            );
+          }
           if (verdictResponse.verdict === Verdict.UNKNOWN) {
```
lennartdohmann, line 261: It's double checked in line 246 but doesn't matter from a functional point of view so idc

unglaublicherdude:
The diff does not seem very helpful, I recommend you to look at the function without the diff.

PR #428, opened 2024-04-09
Tag tests that produces warnings or error logs

lennartdohmann:


unglaublicherdude:
do you also want to add a task in gradle, where you can ignore these test?

Something like

```java
test {
    useJUnitPlatform {
        excludeTags "ErrorLogProducer"
    }
    testLogging {
      events "passed", "skipped", "failed"

      showExceptions true
      exceptionFormat "full"
      showCauses true
      showStackTraces true
    }
}
```

.gitignore
```diff
@@ -56,3 +56,5 @@ _trial_temp*/
 ruby/.idea
 ruby/Gemfile.lock
 ruby/test/test.txt
+
+.env
```
doxthree, line 60: I prefer this, since we have different varients of env files
```suggestion
.env*
```

.gitignore
```diff
@@ -15,7 +15,7 @@ vendor/
 *.cache
 
 # local env files
-.env*
```
GermanCoding, line 18: Shouldn't this already be covered by `.env*` in line 60? AFAIK git matches subdirectories by default

PR #430, opened 2024-04-09
Add authority check for forUrl

lennartdohmann:
Add authority check for forUrl to get a VaasClientException if it is a client side mistake

### DIFF
--- a/java/src/main/java/de/gdata/vaas/ClientCredentialsGrantAuthenticator.java
+++ b/java/src/main/java/de/gdata/vaas/ClientCredentialsGrantAuthenticator.java
@@ -3,6 +3,7 @@ package de.gdata.vaas;
 import com.google.gson.JsonParser;
 import lombok.Getter;
 import lombok.NonNull;
+import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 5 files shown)


unglaublicherdude:
Hm. Wouldn't a toUri be more effective here? Have not tested if that would fail.

unglaublicherdude:
https://www.baeldung.com/java-validate-url

Just want to avoid that we build our own url validation

unglaublicherdude:
Yeah. To Uri should work.

```java
    public void testing() {
        Exception exception = assertThrows(URISyntaxException.class, () -> {
            var url_1 = new URL("https://").toURI();
        });
    }
```

GermanCoding:
> A different question is: Why not just accept URIs instead of URLs as parameter.

I don't think this makes sense. For example, this

`urn:oasis:names:specification:docbook:dtd:xml:4.1.2`

is a valid URI as per RFC 3986 (it's a URN, which is also a type of URI). However, it is not a URL. URIs are supersets of URLs and include much more than just URLs. However, we only want to support URLs (and not even all types of URLs, e.g. not file URLs)

unglaublicherdude:
But than at least we are not consistent 

https://github.com/GDATASoftwareAG/vaas/blob/main/java/src/main/java/de/gdata/vaas/ClientCredentialsGrantAuthenticator.java#L32

lennartdohmann:
But the [marked usage of type URI](https://github.com/GDATASoftwareAG/vaas/blob/main/java/src/main/java/de/gdata/vaas/ClientCredentialsGrantAuthenticator.java#L32) makes kind of sense there because a few lines blow the [HTTPRequest builder](https://github.com/GDATASoftwareAG/vaas/blob/e78d17d1d3d7b85b0c0973e74e81b9a5c8d06d17/java/src/main/java/de/gdata/vaas/ClientCredentialsGrantAuthenticator.java#L63) expects an URI. Otherwise you would create a URI there, convert it to URL and then convert it back to the supertype URI for use.

PR #447, opened 2024-05-05
it is better to use the repository field to link to the repository

szabgab:


GermanCoding (approved):
Failing CI is to be expected since outside contributors do not have access to our secrets in CI.

The changes are fine and do seem to match best pratices (https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field)

PR #490, opened 2024-06-26
set a go-compatible version as tag

unglaublicherdude:


unglaublicherdude:
Out golang versions are not really respected in our current format.

You can see that in the post to the go package system:
image

And also in our listing on the [go package system itself](https://pkg.go.dev/github.com/GDATASoftwareAG/vaas/golang/vaas/pkg/vaas?tab=versions):
image

unglaublicherdude:
With my change we can now tag a new version like 
v1.2.3-golang

Which should be a compatible version, that will also be shown on the golang package system. It is still not perfect, because we would have to keep the -golang suffix in there, so it might get marked as "prerelease" [as described here](https://go.dev/doc/modules/version-numbers#pre-release-version).

image


I don't know how that will look when doing a `go get` but I guess our current output when installing vaas isn't perfect either.

image

unglaublicherdude:
Ah. And I dunno if go get without a version will install the versions with a "prerelease" string

unglaublicherdude:
[Ok, it will not](https://go.dev/doc/modules/release-workflow).

`When making a pre-release available, keep in mind that developers using the pre-release will need to explicitly specify it by version with the go get command. That’s because, by default, the go command prefers release versions over pre-release versions when locating the module you’re asking for. So developers must get the pre-release by specifying it explicitly, as in the following example:`

unglaublicherdude:
This might help.

[Mapping versions to commits](https://go.dev/ref/mod#vcs-version)

PR #493, opened 2024-06-29
Php ping on upload

unglaublicherdude:


php/src/vaas/Vaas.php
```diff
@@ -39,20 +45,17 @@ class Vaas
     private int $_waitTimeoutInSeconds = 600;
     private int $_uploadTimeoutInSeconds = 600;
     private LoggerInterface $_logger;
-    private HttpClient $_httpClient;
     private VaasOptions $_options;
+    private \React\Http\Browser $_httpClient;
 
     /**
      */
     public function __construct(?string $vaasUrl, ?LoggerInterface $logger = null, VaasOptions $options = new VaasOptions())
```
pstadermann, line 53:
```suggestion
    public function __construct(?string $vaasUrl, ?LoggerInterface $logger = new NullLogger(), VaasOptions $options = new VaasOptions())
```

php/src/vaas/Vaas.php
```diff
@@ -63,21 +66,21 @@ public function Connect(
...
         $authRequest = new AuthRequest($token);
         $webSocket->send(json_encode($authRequest));
         $authResponse = $this->_waitForAuthResponse();
-        if ($this->_logger != null)
-            $this->_logger->debug("Authenticated: " . json_encode($authResponse));
+        $this->log("debug", "Authenticated: " . json_encode($authResponse));
         $this->_vaasConnection->SessionId = $authResponse->session_id;
     }
 
+    private function log($level, string|Stringable $message, array $context = []): void {
```
pstadermann, line 79: Mit dem NullLogger kann das raus:

http://phergie.github.io/api/classes/Psr.Log.NullLogger.html



This Logger can be used to avoid conditional log calls

Logging should always be optional, and if no logger is provided to your library creating a NullLogger instance to have something to throw logs at is a good way to avoid littering your code with if ($this->logger) { } blocks.

php/src/vaas/Vaas.php
```diff
@@ -497,14 +492,41 @@ public function setUploadTimeout(int $UploadTimeoutInSeconds): self
...
+            $timeElapsed = time() - $startTime;
+            if ($timeElapsedSinceLastTimestamp >= 5) {
+                $lastTimestamp = time();
+                $this->log("debug", "elapsed time: " . $timeElapsed);
+                $websocket = $this->_vaasConnection->GetAuthenticatedWebsocket();
+                $websocket->ping();
+            }
+        });
+
+        $timeoutTimer = LOOP::addTimer($this->_uploadTimeoutInSeconds, function () {
```
pstadermann, line 510: Maybe use

$browser = $browser->withTimeout(10.0);

https://reactphp.org/http/#timeouts

pstadermann (changes requested):
Very nice. IMHO mergable. Two small things:

Logging: I'd like us to use the NullLogger pattern everywhere.
HTTP timeout: withTimeout is more developer friendly IMHO.

php/src/vaas/Vaas.php
unglaublicherdude: yeah, I had the problem, that it did not call the loop when having nothing in it. so it behaved differently. but I will give it a new try, since I tested it before using the await() function

unglaublicherdude: Ok. Seens to work with the withTimout now

doxthree (approved):

unglaublicherdude:
Replaced by https://github.com/GDATASoftwareAG/vaas/pull/500

PR #499, opened 2024-07-01
Golang: Make VaaS client closeable

rageagainsthepc:
Right now it is not possible to cancel the socket initialization because the context that is passed via `Connect()` is also used for closing the socket. It is more idiomatic in golang to let the `VaaS` client implement the `io.Closer` interface and use that to close the connection explicitly.

### DIFF
--- a/golang/vaas/pkg/vaas/vaas.go
+++ b/golang/vaas/pkg/vaas/vaas.go
@@ -29,7 +29,8 @@ import (
 // The Connect() function has to be called before any other requests are made.
 // Please refer to the individual function comments for more details on their usage and behavior.
 type Vaas interface {
-	Connect(ctx context.Context, auth authenticator.Authenticator) (termChan <-chan error, err error)
+	io.Closer
+	Connect(ctx context.Context, auth authenticator.Authenticator) (errorChan <-chan error, err error)
 	ForUrl(ctx context.Context, uri string) (msg.VaasVerdict, error)
 	ForStream(ctx context.Context, stream io.Reader, contentLength int64) (msg.VaasVerdict, error)
 	ForSha256(ctx context.Context, sha256 string) (msg.VaasVerdict, error)
@@ -59,6 +60,7 @@ var (
 // vaas provides the implementation of the Vaas interface.
 type vaas struct {
 	logger              *log.Logger
+	termChan            chan bool
 	websocketConnection *websocket.Conn
 	openRequests        map[string]chan msg.VerdictResponse
 	requestChannel      chan msg.VerdictRequest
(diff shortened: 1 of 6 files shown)


unglaublicherdude (approved):

golang/vaas/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GDATASoftwareAG/vaas/golang/vaas
 
-go 1.21
```
unglaublicherdude, line 3: Please let that stay on 1.21 as this was requested by the outlook team. the problem is, that in the workflow the go-setup does not use the latest version of 1.21 I will have a look at this.

golang/vaas/pkg/vaas/vaas.go
```diff
@@ -623,36 +635,26 @@ func (v *vaas) runPingTicker(ctx context.Context) {
...
 
-	go func() {
-		<-listenCtx.Done()
-		if err := v.websocketConnection.Close(); err != nil && v.options.EnableLogs {
-			v.logger.Printf("Failed to close web socket: %v", err)
-		}
-		close(termChan)
-	}()
+	go v.runPingTicker()
+	go v.readWebSocket(errorChan)
```
cavus700, line 649: I would prefer to move the go routine inside `readWebSocket` for two reasons:
1. If you do it like this you always have to start the function in a go routing, otherwise the call blocks and you can not read from the channel
2. Not sure if this is an "offical idiom" but for me the one who creates the channel should always be the one who closes it, too. When you pass a channel to a function and the function closes the channel it is some unexpected behaviour and can cause crashes.

```suggestion
	errorChan := v.readWebSocket()
```

cavus700:
Keep in mind that this is a breaking change and will not crash after a simple update in other projects. Furthermore the golang version of vaas does not have correct golang versioning it is unlikely any one will notice it during an update. I don't know how many stakeholders this project has and how you deal with breaking changes but I would not merge this even if the code looks good to me.

unglaublicherdude:
@cavus700 the [Versioning was fixed](https://github.com/GDATASoftwareAG/vaas/pull/491).

cavus700:
Nice, I didn't know about the versioning but in this case this changes would require a v2 for me.

golang/vaas/pkg/vaas/vaas.go
rageagainsthepc: regarding `2.`: So you are basically saying go routins should be anonymous?

rageagainsthepc: I tried to alter as little as possible when it comes to design choices. I would leave this up to @unglaublicherdude if that's ok. Feel free to alter this PR if you think this is a sensible suggestion.

unglaublicherdude: I don't have a heavy opinion here, because we don't really use this SDK in bigger projects. So while you are the ones actually using the SDK I will let you make the decision. But that's just me, maybe @GermanCoding @pstadermann @doxthree or @lennartdohmann have a different opinion on that.

rageagainsthepc: I just went ahead and implemented the requested changes in order to avoid further discussion and get this merged quickly. ;)

unglaublicherdude: @cavus700 @rageagainsthepc do you agree on this PR now? If so, I don't mind merging it.

cavus700: The code itself looks good to me but in my opinion these changes still require a v2 and I would not merge it without. 
Since this is not my project and I don't know all stakeholders the decision is up to you.

unglaublicherdude:
And again I forgot that you have to put the v2 stuff into its own folder.
@rageagainsthepc will you do that or should I? Its go specific, but on Major releases it requires to have the new major release in its own folder. 

So the changes you made have to be in the folder golang/vaas/v2 wich should also be an own module.

An example is the [go-gax](https://github.com/googleapis/gax-go) repository.

PR #500, opened 2024-07-02
replace guzzle with amphp/http-client and revolt/event-loop

unglaublicherdude:


unglaublicherdude:
The order matters, when using the react/event-loop because one http-request can consume the readstream that accidental also in the loop. [See this issue for that](https://github.com/reactphp/http/issues/510).

For a better understanding [I added this test](https://github.com/GDATASoftwareAG/vaas/blob/replace-guzzle-with-react-http/php/tests/vaas/StreamsInLoopTest.php#L22)

What you see is, that with the second http request in a second browser the first stream is getting consumed even though the stream is completely unrelated to the second browser.

unglaublicherdude:
I also opened [this discussion here](https://github.com/GDATASoftwareAG/vaas/blob/replace-guzzle-with-react-http/php/tests/vaas/StreamsInLoopTest.php#L22).

unglaublicherdude:
I also opened [this issue](https://github.com/reactphp/http/issues/533).

unglaublicherdude:
I completely replaced react with amphp and revolt which works as expected. Streams are not getting touched before something is actually consuming them.

unglaublicherdude:
Dunno why it did not start the pipeline. [Here is manual run of these changes](https://github.com/GDATASoftwareAG/vaas/actions/runs/9784912195).

.github/workflows/image-retention.yaml
```diff
@@ -7,6 +7,9 @@ on:
     branches:
       - "*"
       - "!main" # excludes main
+    paths:
```
GermanCoding, line 10: Thanks, that was much needed. Do we still need the branch filter to exclude main with the path filter set?

php/src/vaas/Vaas.php
```diff
@@ -63,18 +68,13 @@ public function Connect(
         string $token,
         ?VaasConnection $vaasConnection = null
     ) {
-        if (isset($vaasConnection)) {
-            $this->_vaasConnection = $vaasConnection;
-        } else {
-            $this->_vaasConnection = new VaasConnection($this->_vaasUrl);
-        }
+        $this->_vaasConnection = isset($vaasConnection) ? $vaasConnection :  new VaasConnection($this->_vaasUrl);
```
GermanCoding, line 71: We can use the null coalescing operator here:

```suggestion
        $this->_vaasConnection = $vaasConnection ?? new VaasConnection($this->_vaasUrl);
```

php/src/vaas/Vaas.php
```diff
@@ -190,15 +183,12 @@ public function ForFile(string $path, $upload = true, string $uuid = null): Vaas
...
     {
-        if ($uuid == null) {
-            $uuid = UuidV4::getFactory()->uuid4()->toString();
-        }
-
-        $verdictResponse = $this->_verdictResponseForStream(
-            $uuid
-        );
+        $this->_logger->debug("uuid: ".var_export($uuid, true));
+        $uuid = $uuid !== null ? $uuid : UuidV4::getFactory()->uuid4()->toString();
```
GermanCoding, line 189:
```suggestion
        $uuid = $uuid ?? UuidV4::getFactory()->uuid4()->toString();
```

php/src/vaas/Vaas.php
```diff
@@ -190,15 +183,12 @@ public function ForFile(string $path, $upload = true, string $uuid = null): Vaas
...
+    public function ForStream(ReadableStream $stream, int $size = 0, string $uuid = null): VaasVerdict
     {
-        if ($uuid == null) {
-            $uuid = UuidV4::getFactory()->uuid4()->toString();
-        }
-
-        $verdictResponse = $this->_verdictResponseForStream(
-            $uuid
-        );
+        $this->_logger->debug("uuid: ".var_export($uuid, true));
```
GermanCoding, line 188: Considering that uuid is typed to be a string, isn't the var_export a bit too much here? A simple . $uuid should suffice I think.

php/src/vaas/Vaas.php
```diff
@@ -190,15 +183,12 @@ public function ForFile(string $path, $upload = true, string $uuid = null): Vaas
...
+        $uuid = $uuid !== null ? $uuid : UuidV4::getFactory()->uuid4()->toString();
+        $this->_logger->debug("uuid: ".var_export($uuid, true));
```
GermanCoding, line 190: Double debug print looks like this is a leftover from debugging. A single print in this line is enough debug logging IMHO.

php/tests/vaas/VaasTest.php
```diff
@@ -64,9 +77,9 @@ public function setUp(): void
         }
     }
 
-    private function _getVaas(bool $useCache = false, bool $useHashLookup = true): Vaas
+    private function _getVaas(bool $useCache = false, bool $useHashLookup = true, LoggerInterface $logger = null): Vaas
     {
-        return new Vaas($_ENV["VAAS_URL"], $this->_getDebugLogger(), new VaasOptions($useCache, $useHashLookup));
+        return new Vaas($_ENV["VAAS_URL"], $logger != null ? $logger : $this->_getDebugLogger(), new VaasOptions($useCache, $useHashLookup));
```
GermanCoding, line 82:
```suggestion
        return new Vaas($_ENV["VAAS_URL"], $logger ?? $this->_getDebugLogger(), new VaasOptions($useCache, $useHashLookup));
```

.github/workflows/image-retention.yaml
unglaublicherdude: Nice hint. I'll tried it out. When no branch-filter matches the pipeline does not run at all.

php/src/vaas/Vaas.php
unglaublicherdude: Both where debugging stuff before I found out, that in VSCode you just have to start the debugger first, then you will basically debug anything you run in php like starting the test from the testexplorer but also from console. That made my life so much easier.

php/tests/vaas/VaasTest.php
```diff
@@ -557,29 +623,59 @@ public function testForStream_WithCleanUrlContentAsStream_ReturnsClean()
...
-
-        $verdict = $vaas->ForStream($stream);
+        $verdict = $vaas->ForStream($body, $size);
 
         $this->assertEquals(Verdict::MALICIOUS, $verdict->Verdict);
         $this->assertEquals("text/plain", $verdict->MimeType);
         $this->assertNotEmpty($verdict->Detection);
     }
+
+    static function random_strings($length_of_string) {
```
doxthree, line 668: There is a duplicate function here. https://github.com/GDATASoftwareAG/vaas/blob/48121e033dae91af41fb3b3b43e122175133b97a/php/tests/vaas/StreamsInLoopTest.php#L29

php/src/vaas/Authentication/OAuth2TokenReceiver.php
```diff
@@ -0,0 +1,69 @@
...
+                break;
+            case "client_credentials":
+                $this->_formParams->addField('client_secret', $this->_clientSecret);
+                break;
+            default:
+                throw new VaasAuthenticationException("Invalid grant type");
+        }
+    }
+
+    public function GetToken() {
```
doxthree, line 53: I'm a php noob so im not familier with naming convention but should the function be called this? 
```suggestion
    public function getToken() {
```

unglaublicherdude: Yeah you are right. At least it is the casing, the two other classes use.

php/tests/vaas/VaasTest.php
unglaublicherdude: Ah yes. I had problems calling stuff from another class and just copied it here. And in the end, I don't need it anymore. Will remove it.

doxthree (approved):

php/tests/vaas/VaasTest.php
```diff
@@ -531,15 +539,73 @@ public function testForStream_WithCleanString_ReturnsClean()
...
-        $httpClient = new Client();
-        $response = $httpClient->get($url);
-        $stream = new Stream($response->getBody()->detach());
+        $verdict = $vaas->ForStream($body, $size);
+        $this->assertEquals(Verdict::CLEAN, $verdict->Verdict);
+    }
 
-        $verdict = $vaas->ForStream($stream);
+    /**
+     * @throws GuzzleException
```
pstadermann, line 562: This seems corrupted. Guzzle is completely removed (?)

pstadermann (approved):
Very nice! Only one comment from me.

PR #520, opened 2024-07-16
Ruby: Fix tests

GermanCoding:


pstadermann (approved):

ruby/test/vaas_test.rb
```diff
@@ -172,7 +175,7 @@ def create(token = nil, timeout = nil)
 
       specify 'upload_failed' do
         vaas, token = create
-        message = {"url" => "https://upload-vaas.gdatasecurity.de/upload", "upload_token" => "invalid_token"}
+        message = {"url" => "https://upload.staging.vaas.gdatasecurity.de/upload", "upload_token" => "invalid_token"}
```
doxthree, line 178: why staging?

GermanCoding: No reason, any env would work here. Ideally this would automatically choose the correct env in each test run, but I didn't bother.

lennartdohmann (approved):

PR #544, opened 2024-07-30
Php connection gets lost when idle after connect

unglaublicherdude:
This is not complete yet. Missing is the Exceptions throw in the EventLoop while no request is running.


unglaublicherdude:
Ok. This is the first version, where all tests pass.

lennartdohmann (approved):

unglaublicherdude:
With the update to "amphp/file": "3.1.1" this can now be reviewed.

unglaublicherdude:
The failing run is the Delete old container images and I am not sure, why that runs anyway within this pr.

unglaublicherdude:
Ok. I fixed the delete container images run but I am still not sure, why it runs in this PR (and other PRs aswell)

unglaublicherdude:
I still see one major issue. Illuminate/Container does only support PHP >= 8.2. So I will have to downgrade it to a version that does still support 8.1

unglaublicherdude:
my fault. illuminate/containers is not used and was only in my composer.lock file

unglaublicherdude:
replaced by https://github.com/GDATASoftwareAG/vaas/pull/592

PR #545, opened 2024-08-01
Add Content-Length header to file upload request

lennartdohmann:
Add Content-Length header to file upload request


unglaublicherdude:
I am pretty sure we removed that for a reason. But If it works now I am fine with the change

unglaublicherdude (approved):

Issue #558, opened 2024-08-15
all: Disable HTTP/2


GermanCoding:
We are encountering performance issues and sometimes HTTP state errors when HTTP/2 is in use. Because we have no actual benefit for HTTP/2, disabling it in all SDKs seems like the most reliable solution for now.

If there is a need for HTTP/2 in the future, we may revisit this decision.

- [x] Typescript
- [x] Go
- [x] Dotnet
- [x] Python
- [x] Java
- [x] Rust
- [x] Ruby
- [x] PHP

GermanCoding:
PRs are open

PR #562, opened 2024-08-15
dotnet: Disable HTTP/2 in file + stream uploads

GermanCoding:
Also consolidate flows for UploadFile and UploadStream

Part of #558

### DIFF
--- a/dotnet/Vaas/src/Vaas/Vaas.cs
+++ b/dotnet/Vaas/src/Vaas/Vaas.cs
@@ -216,7 +216,7 @@ public class Vaas : IDisposable, IVaas
 
         var response = WaitForResponseAsync(verdictResponse.Guid);
         await UploadStream(stream, verdictResponse.Url, verdictResponse.UploadToken, cancellationToken);
-        
+
         return new VaasVerdict(await response);
     }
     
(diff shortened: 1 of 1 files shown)


dotnet/Vaas/src/Vaas/Vaas.cs
```diff
@@ -216,14 +216,16 @@ public async Task<VaasVerdict> ForStreamAsync(
...
-        
+
         return new VaasVerdict(await response);
     }
     
     private async Task UploadStream(Stream stream, string url, string token, CancellationToken cancellationToken)
     {
         using var requestContent = new StreamContent(stream);
         using var requestMessage = new HttpRequestMessage(HttpMethod.Put, url);
+        requestMessage.Version = HttpVersion.Version11;
```
lennartdohmann, line 227: Isn't that the default anyway according to https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.defaultrequestversion?view=net-8.0

lennartdohmann (approved):

dotnet/Vaas/src/Vaas/Vaas.cs
doxthree: It's set by default but in case there are lib changes, where they changed the version from 1.1 to 2.0, it's better to set it in the SDK. Also, it's more transparent for us.

PR #563, opened 2024-08-15
make the example compatible to the current api

unglaublicherdude:


dotnet/Vaas/src/Vaas/VaasFactory.cs
```diff
@@ -15,7 +16,11 @@ public static IVaas Create(VaasOptions vaasOptions)
 
         var bearerTokenHandler = new BearerTokenHandler(authenticator);
         bearerTokenHandler.InnerHandler = new HttpClientHandler();
-        var httpClient = new HttpClient(bearerTokenHandler);
+        var httpClient = new HttpClient(bearerTokenHandler)
+        {
+            DefaultRequestVersion = HttpVersion.Version20,
```
lennartdohmann, line 21: I thought we were now only explicitly using version 1?

lennartdohmann: Defaults to it... https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.defaultrequestversion?view=net-8.0

GermanCoding: This is probably a debugging leftover and should be removed entirely, or explicitly set to HTTP 1.1. In any case, the requests we send also override the default policy.

PR #592, opened 2024-09-04
Phpdev async with amphp

unglaublicherdude:
So basically you can now require a specific branch.

Branches that are named like phpdev-* are pushed without the phpdev- prefix to the vaas-php repository.

So when you have a branch called phpdev-somethinggreat you can install the branch with
```bash
composer require gdata/vaas:dev-somethinggreat
```


unglaublicherdude:
This PR also replaces this https://github.com/GDATASoftwareAG/vaas/pull/544

unglaublicherdude:
You can see the asnyc version running in this https://github.com/GDATASoftwareAG/nextcloud-gdata-antivirus/pull/117

unglaublicherdude:
This has to be a major release because there is a big api change

php/examples/VaasExample/GetVerdictByFile.php
```diff
@@ -13,9 +13,9 @@
...
 );
 
-$vaas = new Vaas(
-    getenv("VAAS_URL") ?? "wss://gateway.production.vaas.gdatasecurity.de"
-);
+$vaas = (new Vaas())
+    ->WithUrl(getenv("VAAS_URL") ?? "wss://gateway.production.vaas.gdatasecurity.de")
+    ->build();
 
 $vaas->Connect($authenticator->getToken());
```
unglaublicherdude, line 20: Just to be clear. With the new implementation the Connect is not necessary anymore. Maybe I sould also give an example on how to use it without you having to call connect yourself.

unglaublicherdude: If fixed that in my new commit.

php/src/vaas/VaasConnection.php
pstadermann, line 1: Functions should camelCased

php/src/vaas/Vaas.php
pstadermann, line 1: Methods should be camelCase. E.g. the builder uses .With*(), but .build().

pstadermann (approved):

PR #600, opened 2024-09-10
599 switch from chai to jest

doxthree:
- Removed chai and mocha as dep
- Add Jest as new dep
- Apply new testing scheme 

NOTE:
Jest also warns about open handles when running:
```
jest --forceExit --detectOpenHandles

Jest has detected the following 3 open handles potentially keeping Jest from exiting:
```

For this case, I would skip this but we should have a look at this in the future.

### DIFF
--- a/typescript/tests/test.ts
+++ b/typescript/tests/test.ts
@@ -1,5 +1,3 @@
-import { expect } from "chai";
-import { describe } from "mocha";
 import * as dotenv from "dotenv";
 import { Vaas } from "../src/Vaas";
 import * as randomBytes from "random-bytes";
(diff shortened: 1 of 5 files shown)


unglaublicherdude:
Nice. I like it. Thank you for putting in the work!

unglaublicherdude (approved):

typescript/tests/test.ts
```diff
@@ -76,108 +72,116 @@ const eicarSha256 =
...
 
-  it("if wrong authentication token is send, an error is expected", async () => {
+  test("if wrong authentication token is send, an error is expected", async () => {
     const token = "ThisIsAnInvalidToken";
     const vaas = new Vaas();
     await expect(
-      (async () => await vaas.connect(token, VAAS_URL))(),
-    ).to.be.rejectedWith("Vaas authentication failed");
+        (async () => await vaas.connect(token, VAAS_URL))(),
+    ).rejects.toThrow("Vaas authentication failed")
```
doxthree, line 85:
```suggestion
    ).rejects.toThrow("Vaas authentication failed");
```

PR #633, opened 2024-10-18
Add common build tool

secana:
This PR adds `just` as a command runner and a `nix` development shell with all tools installed. This allows for easy build, test and release of all SDKs without the need to search through the code how each individual SDK is built. It works on all Linux and macOS (Unix) platforms,  where Nix is supported.

This should significantly reduce the friction for contributing to the project and provides a documentation to build and test the SDKs in code.

ATTENTION:
The `release-*` commands are untested, as I didn't want to release a bunch of SDKs to test the commands. If you release a new SDK version, please check if the command works and fix it if not.

I fumbled with the C++ and Ruby environment variable handling for the tests. It could be possible that the CI pipeline breaks and the SDKs need a fix, as I couldn't test that before.


Readme.md
```diff
@@ -63,27 +63,58 @@ We provide SDKs for various programming languages to make it easy for you to int
...
+|Custom Guids for tracability on user side|&#10060;|&#10060;|&#9989;|&#10060;|&#10060;|&#9989;|&#10060;|&#10060;|&#10060;|
 
 
 ## Integration Ideas for Malware Detection trough VaaS
 You can use VaaS to create various applications that scan for malicious content with a few lines of code. Here are some examples:
 
-Create a command line scanner to find malware: [Example](rust/examples/gscan)
-<img src="assets/gscan.gif" alt="GScan command line malware scanner" style="width:100%">
+- [WordPress Plugin](https://wordpress.org/plugins/gdata-antivirus/) to scan for malware in uploaded files
+- [Nextcloud App](https://apps.nextcloud.com/apps/gdatavaas) to scan files in your Nextcloud instance
```
unglaublicherdude, line 84: thank you for adding those!

cpp/dotenv.h
unglaublicherdude, line 1: nice

Readme.md
```diff
@@ -63,27 +63,58 @@ We provide SDKs for various programming languages to make it easy for you to int
...
+just -l # list all available tasks
+
+# Just Examples
+# Run the tests for the Rust SDK
+just test-rust
+```
+
+There are `test-*`, `build-*`, `clean-*` and `release-*` tasks for each SDK, with the exception of Python and PHP, where no build task is available. You can also run the tests for all SDKs with `just test-all`. A `build-all` and `clean-all` task is available as well.
+
+The `release-*` task triggers a Github Action to build and release a new version of the specified SDK. It needs a version number as an argument, which is used to tag the release. The version number should follow the [Semantic Versioning](https://semver.org/) scheme.
```
unglaublicherdude, line 107: does this actually create a github release?

Readme.md
```diff
@@ -63,27 +63,58 @@ We provide SDKs for various programming languages to make it easy for you to int
...
+
+```bash
+# Example: Release the Rust SDK with version 0.1.0
+just release-rust 0.1.0
+```
+
+As the SDKs need credentials to authenticate against the VaaS API, you need to provide them in a `.env` file. Copy your `.env` file in the root of the project. The C++ SDK needs special credentials, which need to be provided in the `.cpp.env` file.
```
unglaublicherdude, line 114:
```suggestion
As the SDKs need credentials to authenticate to the VaaS API. You need to provide them in a `.env` file. Copy your `.env` file into the root directory of the project. The C++ SDK needs special credentials, which you can provide in a `.cpp.env` file.
```

unglaublicherdude: Maybe an Example of the .env-Files would be great.

unglaublicherdude (approved):

Readme.md
secana: It's just a shortcut for `git tag -a v1.2.3 -m "Release ..." && git push origin v1.2.3`. That's how I released new versions of the SDKs. If a Github release is created is up to the CI pipeline, which I did not touch. So I'm not sure if the pipeline creates a release for the artifacts on push.

unglaublicherdude: Ah I see. We actually do it the other way around. Meaning we create a github-release in the UI and with the release we create the tag. We could discuss doing it your way and implement a github-release step into the pipelines.

cpp/dotenv.h
```diff
@@ -0,0 +1,84 @@
...
+        
+        if (!file) {
+            return env;
+        }
+        
+        for (std::string line; std::getline(file, line);) {
+            const auto pos = line.find('=');
+            if (pos != std::string::npos) {
+                const auto key = line.substr(0, pos);
+                auto value = line.substr(pos + 1);
```
GermanCoding, line 44: Might throw an out_of_range exception if the line ends with =. Do we care?

cpp/dotenv.h
```diff
@@ -0,0 +1,84 @@
...
+
+                env[key] = value;
+            }
+        }
+
+        return env;
+    }
+
+    static void removeCharsFromString(std::string& str, char* charsToRemove) {
+        for (unsigned int i = 0; i < strlen(charsToRemove); ++i) {
```
GermanCoding, line 56: `strlen()` scans for the terminating NULL-byte in the string to find its length. But above we define a simple char-array that doesn't fully represent a string and as such also doesn't contain a NULL-byte. We probably get lucky here because the stack likely has a lot of zeros that eventually cause us to terminate.

```suggestion
        for (unsigned int i = 0; i < sizeof(charsToRemove); ++i) {
```

cpp/dotenv.h
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+  public:
+    Dotenv() : envFile(".env") {
+        this->envFromFile = readEnvFromFile();
+    };
+
+    Dotenv(const std::string& envFile) : envFile(envFile) {
+        this->envFromFile = readEnvFromFile();
+    };
```
GermanCoding, line 68:
```suggestion
    Dotenv() : Dotenv(".env") {}

    Dotenv(const std::string& envFile) : envFile(envFile) {
        this->envFromFile = readEnvFromFile();
    }
```

GermanCoding (changes requested):

flake.nix
```diff
@@ -0,0 +1,105 @@
+{
+  description = "A flake for easy development of the multi-language Verdict-as-a-Service librarie";
```
GermanCoding, line 2:
```suggestion
  description = "A flake for easy development of the multi-language Verdict-as-a-Service libraries";
```

ruby/test/vaas_test.rb
```diff
@@ -172,6 +170,24 @@ def create(token = nil, timeout = nil)
...
+          File.open("test.txt", "w") { |f| f.write(random_text) }
+
+          vaas.connect(token)
+          assert_raises VAAS::VaasUploadError do
+            vaas.upload(message, "./test.txt").wait
+          end
+        ensure
+          vaas.close
+        end
+      end
```
GermanCoding, line 189: We removed the old upload failed test as it didn't have much use. The upload endpoint used here doesn't exist anymore and will get nothing but 404 in any case. That's also why it gets a self-signed cert. I think we can remove this entirely.

justfile
```diff
@@ -0,0 +1,261 @@
...
+alias cgo := clean-go
+
+alias tpy := test-python
+alias cpy := clean-python
+
+alias tph := test-php
+alias cph := clean-php
+
+alias bja := build-java
+alias tja := test-java
```
GermanCoding, line 252: 😄

GermanCoding (changes requested):

GermanCoding:
Very cool!

GermanCoding:
We should also consider enabling Nix support for renovate (https://docs.renovatebot.com/modules/manager/nix/) and probably also consider enabling https://docs.renovatebot.com/configuration-options/#lockfilemaintenance such that the flake.lock gets updated from time to time.

unglaublicherdude:
> We should also consider enabling Nix support for renovate (https://docs.renovatebot.com/modules/manager/nix/) and probably also consider enabling https://docs.renovatebot.com/configuration-options/#lockfilemaintenance such that the flake.lock gets updated from time to time.

I just added a commit, that does that.

Readme.md
secana: Makes sense. But for this PR I would not change anything as so far, both approaches seem to work fine.

cpp/dotenv.h
secana: tbh, I don't care. The parser is only for the tests and not meant to be a fully correct env. parser. It works for what we need, so I would keep it simple.

ruby/test/vaas_test.rb
secana: I see. I didn't get the test running so I decided to ignore it, as I didn't want to touch any of the logic of any of the SDKs. I'll remove the test.

secana:
> I appreciate the work and it looks like a good Idea to have a possbillity to build all the projects with one tool.
> 
> I personally still go with the devcontainers I think, because the devcontainer environments do more than this, they also provide syntax highlighting, auto-complete, a test-explorer, debugging and you can use the remote containers functionality where it basically behaves as you run it locally but it actually runs on some remote system. That's really important for me, doing work from a rather slow internet connection.

You can add any IDE. GUI or terminal, to the flake.nix with a corresponding config. But I didn't want to provide an IDE but a simple tool to test all SDKs with one command in the case that there is a change in the backend. As I don't know all of the languages well, it was hard for me to remember how to build and release new versions of the SDK.

PR #634, opened 2024-10-18
Rust/fix concurrent requests

pstadermann:
Please review carefully, if you time and look out for code that is not idiomatic rust.

### DIFF
--- a/rust/src/response_broker.rs
+++ b/rust/src/response_broker.rs
@@ -69,11 +69,14 @@ mod tests {
     }
 
     #[tokio::test]
-    pub async fn get_response_if_set_all_returns() {
+    pub async fn get_response_if_set_all_returns_response() {
         let responses: ResponseBroker<i32, crate::error::Error> = ResponseBroker::new();
         let response_future = responses.get_response(TEST_REQUEST_ID.to_string());
+
         responses.set_all_responses(Ok(42));
+
         assert_eq!(response_future.await.unwrap(), 42);
+        assert_eq!(responses.responses.lock().unwrap().len(), 0);
     }
 
     #[tokio::test]
--- a/rust/src/response_broker.rs
+++ b/rust/src/response_broker.rs
@@ -6,9 +6,10 @@ use std::sync::Mutex;
 use tokio::sync::oneshot;
 use tokio::sync::oneshot::error::RecvError;
 use tokio::sync::oneshot::Sender;
+use tracing::{error, debug};
 
 #[derive(Debug)]
-pub(crate) struct ResponseBroker<T: Clone + Debug, E: From<RecvError> + Clone + std::error::Error> {
+pub(crate) struct ResponseBroker<T: Clone + Debug, E: std::error::Error + Clone + From<RecvError>> {
     responses: Mutex<HashMap<String, Sender<Result<T, E>>>>,
 }
 
@@ -33,11 +34,10 @@ impl<T: Clone + Debug, E: From<RecvError> + Clone + std::error::Error> ResponseB
         let mut requests = self.responses.lock().unwrap_or_else(|e| e.into_inner());
         if let Some(r) = requests.remove(request_id) {
             if r.send(response).is_err() {
-                // TODO: proper logging
-                eprintln!("cannot send");
+                debug!("Receiver for {request_id} has been dropped");
             }
         } else {
-            eprintln!("cannot find request id");
+            error!("Can't find receiver for {request_id}");
         }
     }
 
@@ -55,6 +55,7 @@ impl<T: Clone + Debug, E: From<RecvError> + Clone + std::error::Error> ResponseB
 
 #[cfg(test)]
 mod tests {
+    use tracing_test::traced_test;
     use super::ResponseBroker;
 
     const TEST_REQUEST_ID: &str = "1234";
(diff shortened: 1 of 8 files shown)


GermanCoding (approved):

secana:
What does it fix? The title says "concurrent request". But from the code I'm not able to see where is issue was that is fixed here.

secana:
Do we want to introduce logging in a library? That's not uncommon in Rust, but I'm not to fond of it, as the library-user has to manually disable/set the tracing for each library that is used. That leads to beautiful code like this: https://github.com/kellnr/kellnr/blob/df94e8894948094776cbe379c9e683cb07cfdf9c/crates/kellnr/src/main.rs#L202

My expectation for a library is that errors are bubbled up as values and than logged by the library users. But that's not a hill I'll die on.

GermanCoding:
The issue fixed here is that the current implementation is prone to data races when multiple concurrent requests are issued. The broadcast channel used doesn't have strong enough ordering guarantees (the order in which we call subscribe does not necessarily match the order in which responses come in), and as such, verdict responses may inadvertently get mapped to incorrect requests. The entire approach is not sound as-is. 

Also, the old channel tends to quickly lose responses when issuing requests at a moderately (~10 req/s) high rate due to the broadcast channel filling up (this used to be configurable, but the new approach fixes the problem altogether).

We cannot bubble up errors within the web socket, as this is code that isn't externally exposed in any way. We only have the choice of either suppressing the errors silently, straight up panicking, or logging them. This PR chooses the latter option so that there's at least some chance of us seeing problems if they do occur. The application can configure the logging if they want.

secana:
Thanks for the explanation. The mapping of the request is done based on the `guid` of the message and not based on the ordering. The ordering can be random. I'm not sure if the issue the PR tries to fix exists in the first place. As far as I can see, `subscribe` is always called before the message is send, so that does not seem to be an issue. But maybe I'm overlooking something. Just want to understand the issue, even if the new code is cleaner and easier to reason about.

I switched from an unbound channel to a bound one as the backend died when too many requests were send. I used the channel capacity as a cheap back-pressure mechanism. If that's not the case anymore, we do not need the safe guard in the client anymore and can switch to an unbound channel, or as in this case a one-shot. Not sure if we run in timeouts if we fully utilize the network connection as we allow infinite parallel requests, but that's easy to test.

I see the difficulty with the error handling, but need more time to think about a possible solution. Maybe there is a smart way to get the errors out, but that does not have take place in this PR.

I like that the PR makes the code a lot cleaner and centralizes the logic for sending and waiting for messages. I'll give it test, when I'm back in the office.

GermanCoding:
The issue is that the order in which subscribe is called is not necessarily related to the order in which requests are uploaded or replied to by the server - if two requests are submitted simultaneously, we call subscribe() in order A, upload in order B, and get verdict responses in order C. There is nothing to ensure that A = B = C. (The order of upload technically doesn't matter, as long as A  = C holds). The new code maps by GUIDs instead, which ensures that the order becomes irrelevant and fully parallel processing is possible.

pstadermann:
What does this fix: The current version of the VaaS rust SDK fails sporadically, if one instance is used concurrently. This becomes apparent with the new gscan release (see next PR).

rust/src/response_broker.rs
```diff
@@ -0,0 +1,89 @@
...
+            receiver
+        };
+        receiver.map(|r| r?)
+    }
+
+    pub fn set_response(&self, request_id: &str, response: Result<T, E>) {
+        let mut requests = self.responses.lock().unwrap_or_else(|e| e.into_inner());
+        if let Some(r) = requests.remove(request_id) {
+            if r.send(response).is_err() {
+                debug!("Receiver for {request_id} has been dropped");
```
secana, line 37: IHMO this should be `error!` not `debug!` as file was not scanned successfully and no verdict is available.

rust/src/response_broker.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+impl<T: Clone + Debug, E: From<RecvError> + Clone + std::error::Error> ResponseBroker<T, E> {
+    pub fn new() -> Self {
+        Self {
+            responses: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub fn get_response(&self, request_id: String) -> impl Future<Output = Result<T, E>> {
```
secana, line 23: NIT: Maybe `get_receiver` or `get_response_receiver` are better names? We are not getting a response back here.

rust/src/response_broker.rs
GermanCoding: This is debug, because it can happen during normal operation - a caller can go away due to timeout.

rust/src/response_broker.rs
```diff
@@ -0,0 +1,89 @@
...
+
+#[derive(Debug)]
+pub(crate) struct ResponseBroker<T: Clone + Debug, E: std::error::Error + Clone + From<RecvError>> {
+    responses: Mutex<HashMap<String, Sender<Result<T, E>>>>,
+}
+
+impl<T: Clone + Debug, E: From<RecvError> + Clone + std::error::Error> ResponseBroker<T, E> {
+    pub fn new() -> Self {
+        Self {
+            responses: Mutex::new(HashMap::new()),
```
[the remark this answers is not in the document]
GermanCoding: This is technically a map that maps GUIDs to their respective response-oneshot-channels. A request isn't really involved here, except for the fact that the GUID is used in both request and response.

pstadermann:
> The oneshot implementation is a lot cooler

We know :)

> than the broadcast with guid matching.

There was no guid matching. The first listener got the response. There was no checking for the request_id.

rust/src/response_broker.rs
GermanCoding: Yeah we also discussed the naming here. This choice was made to "hide" the underlying implementation from the caller. The fact that this method actually uses a channel to receive the response is an implementation detail, not a feature of the function. The function's contract just returns a future that will (eventually) yield a response.

GermanCoding:
To clarify the old code, because it's confusing (and I actually also made a mistake in my earlier explanation): The old code _tried_ to perform GUID matching on the response, but this code never worked: If a receiver got the wrong response, that response was just dropped - it never reached the correct receiver.

PR #643, opened 2024-10-31
fix: don't create a new WebsocketClient object reference for every connection delay tick

HuseinJ:
Initialize WebSocketClient only once to prevent redundant re-creation during connection retries, addressing potential issues with garbage collection when OS-level blocking operations may prevent proper cleanup, leading to excessive thread accumulation.


pstadermann:
Thank you for the PR.

We can reproduce this. I expanded on your PR in https://github.com/GDATASoftwareAG/vaas/issues/647.

We will release a new version this morning.

HuseinJ:
Thank you too for the fast response and implementation time 😄

PR #654, opened 2024-11-15
golang HTTP API

GermanCoding:


golang/vaas/v2/pkg/authenticator/authenticator.go
```diff
@@ -67,12 +68,43 @@ type Authenticator interface {
...
+// cachedToken is a cached access token that may be reused
+type cachedToken struct {
+	accessToken string
+	expires     time.Time
+}
+
+// ShouldRefresh determines whether this token should be replaced by a newer one
+func (c cachedToken) ShouldRefresh() bool {
+	// Suggest refreshing the token slightly early to avoid races
+	return time.Now().Add(time.Minute).After(c.expires)
```
pstadermann, line 90: Should not be necessary. Server adds clock skew.

golang/vaas/v2/pkg/vaas/vaas.go
```diff
@@ -79,29 +129,21 @@ func NewWithDefaultEndpoint(options options.VaasOptions, authenticator authentic
 func (v *vaas) ForSha256(ctx context.Context, sha256 string) (msg.VaasVerdict, error) {
 	// Loop until we get 200 or an error
 	for {
-		token, err := v.authenticator.GetToken()
-		if err != nil {
-			return msg.VaasVerdict{}, err
+		select {
+		case <-ctx.Done():
```
pstadermann, line 153: Do we need this?

GermanCoding: Technically no, but it's generally a bit more user friendly (prevents triggering a HTTP request when the ctx is already cancelled). If we are fine with creating a garbage request (that is immediately going to be thrown away during I/O), we can remove this.

PR #676, opened 2024-12-11
Java/http api

doxthree:


.github/workflows/ci-java.yaml
```diff
@@ -89,28 +89,28 @@ jobs:
...
-          gradle-version: "8.6"
-          build-root-directory: java/examples/VaasExample
-
-      - name: run authentication examples
-        uses: gradle/gradle-build-action@v3
-        with:
-          arguments: authentication
-          gradle-version: "8.6"
-          build-root-directory: java/examples/VaasExample
+      # - name: run examples for file
```
pstadermann, line 92: Reenable or remove

java/src/main/java/de/gdata/vaas/ClientCredentialsGrantAuthenticator.java
pstadermann, line 1: Does not cache and reuse the token. See golang.

java/src/main/java/de/gdata/vaas/IVaas.java
pstadermann, line 1: We need both for* and for*Async methods. Most Java users won't use/be familiar with the async stuff.

java/src/main/java/de/gdata/vaas/ResourceOwnerPasswordGrantAuthenticator.java
pstadermann, line 1: Does not cache and reuse the token. See golang.

java/src/main/java/de/gdata/vaas/VaasConfig.java
```diff
@@ -23,17 +23,11 @@ public class VaasConfig {
...
 
     public VaasConfig() throws URISyntaxException {
-        this(
-                new URI("wss://gateway.production.vaas.gdatasecurity.de"));
+        this(new URI("https://gateway.production.vaas.gdatasecurity.de"));
     }
 
     public VaasConfig(URI url) {
         this.url = url;
         this.PullDelayMs = 100;
```
pstadermann, line 31: PullDelayMs is not in use anymore and can be removed.

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
 
-public class Vaas implements AutoCloseable {
-    private static final int connectionRetryDelayInMs = 1000;
-    private static final int connectionTimeoutInMs = 10000;
+public class Vaas implements IVaas {
+    private static final String userAgent = "Java/9.0.0";
```
pstadermann, line 31: Get version at runtime.

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
-                        return null;
+    private static CompletableFuture<VaasVerdict> sendFileWithRetry(HttpClient httpClient, HttpRequest request) {
+        return httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofString())
+                .thenCompose(handleException(response -> {
+                    switch (response.statusCode()) {
+                        case 200:
+                            var fileReport = FileReport.fromJson(response.body());
+                            return CompletableFuture.completedFuture(VaasVerdict.From(fileReport));
+                        case 201:
+                            return sendFileWithRetry(httpClient, request);
```
pstadermann, line 101: Recursion without limit. This could cause a StackOverflow.

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
+
+    private static CompletableFuture<VaasVerdict> sendUrlWithRetry(HttpClient httpClient, HttpRequest request) {
+        return httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofString())
+                .thenCompose(handleException(response -> {
+                    switch (response.statusCode()) {
+                        case 200:
+                            var urlReport = UrlReport.fromJson(response.body());
+                            return CompletableFuture.completedFuture(VaasVerdict.From(urlReport));
+                        case 201:
+                            return sendUrlWithRetry(httpClient, request);
```
pstadermann, line 116: Recursion without limit. This could cause a StackOverflow.

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
-                    }
-                });
+    public HttpRequest.Builder CreateHttpRequestBuilderWithHeaders(URI uri, String requestId)
+            throws IOException, InterruptedException, VaasAuthenticationException {
+        var token = this.authenticator.getToken();
+        return HttpRequest.newBuilder()
+                .uri(uri)
+                .header("Authorization", "Bearer " + token)
+                .header("User-Agent", userAgent)
+                .header("tracestate", "vaasrequestid=" + requestId);
```
pstadermann, line 130: Only add header if we have a requestId.

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
-    private CompletableFuture<VerdictResponse> forRequest(VerdictRequest verdictRequest)
-            throws VaasConnectionClosedException {
-        var verdictResponse = this.client.waitForVerdict(verdictRequest.getGuid());
-
-        verdictRequest.setSessionId(this.client.getSessionId());
-        try {
-            this.client.send(verdictRequest.toJson());
-        } catch (WebsocketNotConnectedException ignored) {
-            throw new VaasConnectionClosedException();
+        if (options.getVaasRequestId() == null || options.getVaasRequestId().isBlank()) {
```
pstadermann, line 154: Don't send vaasRequestId if == null

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
+    @Override
+    public CompletableFuture<VaasVerdict> forFile(Path file, ForFileOptions options)
+            throws NoSuchAlgorithmException, IOException, InterruptedException, VaasAuthenticationException {
+        var sha256 = new Sha256(file);
+        var forSha256Options = new ForSha256Options(options.isUseCache(), options.isUseHashLookup(),
+                options.getVaasRequestId());
+
+        return forSha256(sha256, forSha256Options)
+                .thenCompose(handleException(vaasVerdict -> {
+                    var verdictWithoutDetection = vaasVerdict.getVerdict() == Verdict.MALICIOUS
```
pstadermann, line 180: This and the following if are strange. Please compare to golang.

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
+                            || (vaasVerdict.getDetection() != null && vaasVerdict.getDetection().isBlank());
+
+                    if (vaasVerdict.getVerdict() != Verdict.UNKNOWN && verdictWithoutDetection
+                            && vaasVerdict.getFileType() != null
+                            && !vaasVerdict.getFileType().isBlank()
+                            && vaasVerdict.getMimeType() != null
+                            && !vaasVerdict.getMimeType().isEmpty()) {
+                        return CompletableFuture.completedFuture(vaasVerdict);
+                    } else {
+                        var inputstream = Files.newInputStream(file, StandardOpenOption.READ);
```
pstadermann, line 191: Close string. https://www.infoq.com/news/2010/08/arm-blocks/

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,278 @@
...
+        return httpClient.sendAsync(postRequest, HttpResponse.BodyHandlers.ofString())
+                .thenCompose(handleException(response -> {
+                    var statusCode = response.statusCode();
+                    if (statusCode < 200 || statusCode >= 300) {
+                        return parseVaasError(response);
+                    }
+                    var urlAnalysisStarted = UrlAnalysisStarted.fromJson(response.body());
+                    var urlsReportUri = this.config.getUrl()
+                            + String.format("/urls/%s/report?", urlAnalysisStarted.getId()) + encodeParams(params);
+                    if (options.getVaasRequestId() == null || options.getVaasRequestId().isBlank()) {
```
pstadermann, line 268: Omit vaasRequestId if == null

pstadermann (changes requested):

java/src/main/java/de/gdata/vaas/Vaas.java
pstadermann, line 1: Cancellation and timeouts are not implemented.

java/src/main/java/de/gdata/vaas/Vaas.java
pstadermann: I checked in the debugger. The stack does not grow. OK for now.

java/src/main/java/de/gdata/vaas/Vaas.java
pstadermann: I checked in the debugger. The stack does not grow. OK for now.

java/src/main/java/de/gdata/vaas/IVaas.java
doxthree: done

.github/workflows/ci-java.yaml
doxthree: Will be enabled when new Version is released

java/src/main/java/de/gdata/vaas/VaasConfig.java
doxthree: Done

java/src/main/java/de/gdata/vaas/Vaas.java
doxthree: Done

java/src/main/java/de/gdata/vaas/Vaas.java
doxthree: Done

java/src/main/java/de/gdata/vaas/VaasConfig.java
doxthree: Done

java/src/main/java/de/gdata/vaas/ClientCredentialsGrantAuthenticator.java
doxthree: Done

java/src/main/java/de/gdata/vaas/Vaas.java
doxthree: Java and Gradle does not support this, has to be implemented in a seperated logic

lennartdohmann (changes requested):
Looks good at all. I prefer the authenticators to behave like the other SDKs and currently the specific options overwrite usage of yet missing global cache and hashlookup flags. I made a PR into this with my suggestions. #700

lennartdohmann (approved):

lennartdohmann (changes requested):
Some changes for big files needed. PR incoming

java/src/main/java/de/gdata/vaas/Vaas.java
```diff
@@ -1,655 +1,544 @@
...
-        return this.forSha256(sha256, UUID.randomUUID(), verdictRequestAttributes);
+    @Override
+    public CompletableFuture<VaasVerdict> forFileAsync(Path file, ForFileOptions options)
+            throws IOException, VaasAuthenticationException, NoSuchAlgorithmException {
+        var sha256 = new Sha256(file);
+        var contentLength = Files.size(file);
+        var forSha256Options = new ForSha256Options(options.isUseCache(), options.isUseHashLookup(),
+                options.getVaasRequestId());
+        return forSha256Async(sha256, forSha256Options)
+                .thenCompose(handleException(vaasVerdict -> {
```
GermanCoding, line 272: It is possible that the server has a failed scan in the cache (from a previous failure), which will be returned as an error during forSha256. This then throws the exception here, but it should actually just be ignored: We can just (re-)upload the file and start a new analysis. In general, forSha256 errors should not cause a failure here, but just proceed with the upload & scan (if that also fails, then throw)

GermanCoding (approved):
Looks good, just the one nit about error handling in forFile

java/src/main/java/de/gdata/vaas/Vaas.java
doxthree: Which exception dou you mean?

GermanCoding: Any VaasClientException/VaasServerException that forSha256 might throw

lennartdohmann (approved):

java/examples/VaasExample/src/main/java/de/gdata/vaasexample/ForFileScan.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.nio.file.Path;
+
+public class ForFileScan {
+    public static void main(String[] args) throws Exception {
+        IAuthenticator authenticator;
+        var env = new Environment();
+        if (env.clientId == null) {
+            if (env.userName == null) {
+                throw new IllegalStateException("Either CLIENT_ID or VAAS_USER_NAME must be set");
```
lennartdohmann, line 15: Since this is an example I would prefer to have it as easy as possible with less code and just use one authenticator instead of it depending on whatever is set in the environment. Also CLIENT_ID is always set even on `Password` flow so this makes not really sense due to `clientId` will always be set.

java/examples/VaasExample/src/main/java/de/gdata/vaasexample/Config.java
```diff
@@ -0,0 +1,34 @@
...
+import java.net.URI;
+import java.nio.file.Path;
+
+public class Config {
+    public static void main(String[] args) throws Exception {
+        IAuthenticator authenticator;
+        var env = new Environment();
+        if (env.clientId == null) {
+            if (env.userName == null) {
+                throw new IllegalStateException("Either CLIENT_ID or VAAS_USER_NAME must be set");
```
lennartdohmann, line 15: Same here

java/examples/VaasExample/src/main/java/de/gdata/vaasexample/Config.java
```diff
@@ -0,0 +1,34 @@
...
+            authenticator = new ClientCredentialsGrantAuthenticator(env.clientId, env.clientSecret, new URI(env.tokenUrl));
+        }
+
+        var timeoutInMs = 5000;
+        var useCache = false;
+        var useHashLookup = false;
+        var config = new VaasConfig(timeoutInMs, useCache, useHashLookup, new URI(env.vaasUrl));
+        var vaas = new Vaas(config, authenticator);
+        var file = Path.of(env.scanPath);
+        var verdict = vaas.forFile(file);
```
lennartdohmann, line 28: I would prefer to separate this with line breaks to show how less code is actually needed to scan a file.

java/examples/VaasExample/src/main/java/de/gdata/vaasexample/ForSha256Scan.java
```diff
@@ -0,0 +1,30 @@
...
+import de.gdata.vaas.authentication.*;
+import java.net.URI;
+
+public class ForSha256Scan {
+    public static void main(String[] args) throws Exception {
+        IAuthenticator authenticator;
+        var env = new Environment();
+        if (env.clientId == null) {
+            if (env.userName == null) {
+                throw new IllegalStateException("Either CLIENT_ID or VAAS_USER_NAME must be set");
```
lennartdohmann, line 13: Same here

java/examples/VaasExample/src/main/java/de/gdata/vaasexample/ForSha256Scan.java
```diff
@@ -0,0 +1,30 @@
...
+            }
+            authenticator = new ResourceOwnerPasswordGrantAuthenticator("vaas-customer", env.userName, env.password, new URI(env.tokenUrl));
+        } else {
+            authenticator = new ClientCredentialsGrantAuthenticator(env.clientId, env.clientSecret, new URI(env.tokenUrl));
+        }
+
+        var config = new VaasConfig(new URI(env.vaasUrl));
+        var vaas = new Vaas(config, authenticator);
+        var sha256 = new Sha256("275a021bbfb6489e54d471899f7db9d1663fc695ec2fe2a2c4538aabf651fd0f");
+        var verdict = vaas.forSha256(sha256);
```
lennartdohmann, line 23: Same here

java/examples/VaasExample/src/main/java/de/gdata/vaasexample/ForStreamScan.java
```diff
@@ -0,0 +1,33 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class ForStreamScan {
+    public static void main(String[] args) throws Exception {
+        IAuthenticator authenticator;
+        var env = new Environment();
+        if (env.clientId == null) {
+            if (env.userName == null) {
+                throw new IllegalStateException("Either CLIENT_ID or VAAS_USER_NAME must be set");
```
lennartdohmann, line 16: Same here

java/examples/VaasExample/src/main/java/de/gdata/vaasexample/ForUrlScan.java
```diff
@@ -0,0 +1,31 @@
...
+import java.net.URI;
+import java.net.URL;
+
+public class ForUrlScan {
+    public static void main(String[] args) throws Exception {
+        IAuthenticator authenticator;
+        var env = new Environment();
+        if (env.clientId == null) {
+            if (env.userName == null) {
+                throw new IllegalStateException("Either CLIENT_ID or VAAS_USER_NAME must be set");
```
lennartdohmann, line 14: Same here

lennartdohmann (changes requested):

java/build.gradle
```diff
@@ -22,6 +22,10 @@ dependencies {
     implementation 'com.google.code.gson:gson:2.11.0'
     implementation 'org.jetbrains:annotations:26.0.1'
     implementation 'io.github.cdimascio:dotenv-java:3.1.0'
+    implementation 'com.ibm.async:asyncutil:0.1.0'
+    implementation 'com.amazonaws:aws-java-sdk-s3:1.12.780'
```
GermanCoding, line 26: This should probably be a `testImplementation` right? We don't use S3 outside of the new test?

PR #678, opened 2024-12-12
Php/http_api

lennartdohmann:


unglaublicherdude:
Hey. I don't know what's better for you. I can also make a proposal PR if you like that more. I would rewrite the Authenticator stuff like that:

```php
enum Suit: string
{
    case CLIENT_CREDENTIALS = 'client_credentials';
    case PASSWORD = 'password';
}

interface IAuthenticatorOptions {
}

class ResourceOwnerPasswordGrantAuthenticatorOptions implements IAuthenticatorOptions {
    public string $grandType = 'password';
    public function __construct(public string $clientId, public string $username, public string $password) {}
}

class ClientCredentialsAuthenticatorOptions implements IAuthenticatorOptions {
    public string $grandType = 'client_credentials';
    public function __construct(public string $clientId, public string $clientSecret) {}
}


class someAuthenticator {
    private function tokenRequestToForm(IAuthenticatorOptions $options): string
    {
        return match(true) {
            $options instanceof ResourceOwnerPasswordGrantAuthenticatorOptions => http_build_query([
                'client_id' => $options->clientId,
                'username' => $options->username ?? throw new InvalidArgumentException(),
                'password' => $options->password ?? throw new InvalidArgumentException(),
                'grant_type' => $options->grandType
            ]),
            $options instanceof ClientCredentialsAuthenticatorOptions => http_build_query([
                'client_id' => $options->clientId,
                'client_secret' => $options->clientSecret ?? throw new InvalidArgumentException(),
                'grant_type' => $options->grandType,
            ])
        };
    }
}
```

With that change you don't need the knowledge for what grandType does need what fields, because the classes are specific.

unglaublicherdude (changes requested):

unglaublicherdude:
This PR is allot of work. Thank you for tackling it!

php/examples/VaasExample/GetVerdictByFile.php
```diff
@@ -7,18 +7,19 @@
...
-    getenv("CLIENT_SECRET"),
-    getenv("TOKEN_URL") ?: "https://account.gdata.de/realms/vaas-production/protocol/openid-connect/token"
+    clientId: getenv("CLIENT_ID"),
+    clientSecret: getenv("CLIENT_SECRET"),
+    tokenUrl: getenv("TOKEN_URL")
 );
 
 $vaas = (new Vaas())
     ->withAuthenticator($authenticator)
-    ->withUrl(getenv("VAAS_URL") ?? "wss://gateway.production.vaas.gdatasecurity.de")
```
pstadermann, line 18: Wo wird die URL jetzt gesetzt?

lennartdohmann: The url is part of the VaasOptions here https://github.com/GDATASoftwareAG/vaas/blob/ac30ae0fcdb8023e5ae38fd2d3f30acce8a864d9/php/src/vaas/Options/VaasOptions.php#L17 and defaults to cloud prod

php/examples/VaasExample/GetVerdictByHash.php
```diff
@@ -3,23 +3,28 @@
...
-    getenv("CLIENT_SECRET"),
-    getenv("TOKEN_URL") ?: "https://account.gdata.de/realms/vaas-production/protocol/openid-connect/token"
+    clientId: getenv("CLIENT_ID"),
+    clientSecret: getenv("CLIENT_SECRET"),
+    tokenUrl: getenv("TOKEN_URL")
 );
+
 $vaas = (new Vaas())
     ->withAuthenticator($authenticator)
-    ->withUrl(getenv("VAAS_URL") ?? "wss://gateway.production.vaas.gdatasecurity.de")
```
pstadermann, line 17: Same here

php/examples/VaasExample/GetVerdictByHash.php
```diff
@@ -3,23 +3,28 @@
...
     ->build();
 
+
 // EICAR
-$vaasVerdict = $vaas->ForSha256("000005c43196142f01d615a67b7da8a53cb0172f8e9317a2ec9a0a39a1da6fe8");
-fwrite(STDOUT, "Verdict for $vaasVerdict->Sha256 is " . $vaasVerdict->Verdict->value . " \n");
-// SOMEFILE
-$vaasVerdict = $vaas->ForSha256("70caea443deb0d0a890468f9ac0a9b1187676ba3e66eb60a722b187107eb1ea8");
-fwrite(STDOUT, "Verdict for $vaasVerdict->Sha256 is " . $vaasVerdict->Verdict->value . " \n");
+$vaasVerdict = $vaas->forSha256Async(Sha256::TryFromString("000005c43196142f01d615a67b7da8a53cb0172f8e9317a2ec9a0a39a1da6fe8"))->await();
```
pstadermann, line 24: EICAR is 275a021bbfb6489e54d471899f7db9d1663fc695ec2fe2a2c4538aabf651fd0f.

Comment does not match.

php/examples/VaasExample/GetVerdictByUrl.php
```diff
@@ -7,19 +7,22 @@
...
-    getenv("CLIENT_SECRET"),
-    getenv("TOKEN_URL") ?: "https://account.gdata.de/realms/vaas-production/protocol/openid-connect/token"
+    clientId: getenv("CLIENT_ID"),
+    clientSecret: getenv("CLIENT_SECRET"),
+    tokenUrl: getenv("TOKEN_URL")
 );
+
 $vaas = (new Vaas())
     ->withAuthenticator($authenticator)
-    ->withUrl(getenv("VAAS_URL") ?? "wss://gateway.production.vaas.gdatasecurity.de")
```
pstadermann, line 17: Same here

php/src/vaas/Authentication/ClientCredentialsGrantAuthenticator.php
```diff
@@ -2,17 +2,45 @@
...
+    /**
+     * Gets the access token asynchronously.
+     * If the token is still valid, it will be returned immediately.
+     * If the token is expired, a new token will be requested.
+     * @param Cancellation|null $cancellation Cancellation token
+     * @return Future Future that resolves to the access token string
+     */
+    public function getTokenAsync(?Cancellation $cancellation = null): Future
+    {
+        return async(function () use ($cancellation) {
```
pstadermann, line 42: return $this->tokenReceiver->getTokenAsync($cancellation);

php/src/vaas/Authentication/ResourceOwnerPasswordGrantAuthenticator.php
```diff
@@ -2,20 +2,48 @@
...
+     * If the token is still valid, it will be returned immediately.
+     * If the token is expired, a new token will be requested.
+     * @param Cancellation|null $cancellation Cancellation token
+     * @return Future Future that resolves to the access token string
      */
-    public function getToken(): string {
-        return $this->_tokenReceiver->GetToken();
+    public function getTokenAsync(?Cancellation $cancellation = null): Future
+    {
+        return async(function () use ($cancellation) {
```
pstadermann, line 45: return $this->tokenReceiver->getTokenAsync($cancellation);

php/src/vaas/Vaas.php
```diff
@@ -2,385 +2,455 @@
...
+                $this->options->vaasUrl,
                 $sha256,
-                $uuid
-            )
-        );
+                json_encode($options->useCache),
+                json_encode($options->useHashLookup
+            ));
+
+            $request = new Request($url, 'GET');
```
pstadermann, line 140: Move into loop?

php/src/vaas/Vaas.php
```diff
@@ -2,385 +2,455 @@
...
+                ]);
+                $sha256 = Sha256::TryFromFile($path);
+                $this->logger->debug("Check if file $path is already known by its SHA256: $sha256");
+                $response = $this->forSha256Async($sha256, $forSha256Options, $cancellation)->await();
+                $isVerdictWithoutDetection = ($response->verdict === 'Malicious' || $response->verdict === 'Pup') && !empty($response->detection);
+                if ($response->verdict !== 'Unknown' && !empty($response->fileType) && !empty($response->mimeType) && !$isVerdictWithoutDetection) {
+                    $this->logger->debug("File $path is already known from cache or G DATA cloud by its SHA256: $sha256");
+                    return $response;
+                }
+                $this->logger->debug("File $path is not known from cache or G DATA cloud by its SHA256: $sha256");
```
pstadermann, line 214: Output message is wrong.

php/src/vaas/Vaas.php
```diff
@@ -2,385 +2,455 @@
...
+     * Add the necessary headers to a request:
+     * - Authorization (Bearer token)
+     * - User-Agent
+     * - tracestate
+     * @param Request $request The request to add headers to
+     * @param string|null $requestId The request ID to add to the headers
+     * @return Future A future that resolves when the headers have been added
      */
-    public function setWebsocketTimeOut(int $timeoutInSeconds): void
+    private function addRequestHeadersAsync(Request $request, ?string $requestId = ''): Future
```
pstadermann, line 399: Why is this Async?

php/src/vaas/Vaas.php
```diff
@@ -2,385 +2,455 @@
...
+                $isVerdictWithoutDetection = ($response->verdict === 'Malicious' || $response->verdict === 'Pup') && !empty($response->detection);
+                if ($response->verdict !== 'Unknown' && !empty($response->fileType) && !empty($response->mimeType) && !$isVerdictWithoutDetection) {
+                    $this->logger->debug("File $path is already known from cache or G DATA cloud by its SHA256: $sha256");
+                    return $response;
+                }
+                $this->logger->debug("File $path is not known from cache or G DATA cloud by its SHA256: $sha256");
+            }
+
+            $stream = openFile($path, 'r');
+            $stream->onClose(fn() => $stream->close());
```
pstadermann, line 218: Does this work?

php/examples/VaasExample/GetVerdictByFile.php
```diff
@@ -7,18 +7,19 @@
...
-    getenv("CLIENT_ID"),
-    getenv("CLIENT_SECRET"),
-    getenv("TOKEN_URL") ?: "https://account.gdata.de/realms/vaas-production/protocol/openid-connect/token"
+    clientId: getenv("CLIENT_ID"),
+    clientSecret: getenv("CLIENT_SECRET"),
+    tokenUrl: getenv("TOKEN_URL")
 );
 
-$vaas = (new Vaas())
+$vaas = Vaas::builder()
```
unglaublicherdude, line 17: Nice. I like that very much.

php/examples/VaasExample/GetVerdictByFile.php
unglaublicherdude: Maybe we should include an example where we also set different VaasOptions.

php/src/vaas/Authentication/TokenResponse.php
unglaublicherdude, line 1: Why is this the only message, that's not in the Message Folder?
Maybe it would be better to have all classes ordered by topic instead of type?

Also I would go with [Constructor property promotion](https://stitcher.io/blog/constructor-promotion-in-php-8).


```php
<?php

namespace VaasSdk\Authentication;

use InvalidArgumentException;

class TokenResponse
{
    public function __construct(public string $accessToken, public ?int $expiresInSeconds = null)
    {
        if (empty($accessToken)) {
            throw new InvalidArgumentException('Access token cannot be null or empty');
        }
    }
}
```

The more I see it, the more I like it and you cannot forget to actually set the properties.

php/src/vaas/Exceptions/FileDoesNotExistException.php
unglaublicherdude, line 1: As the class that extends inherits the constructor of the extended class you can just do this:

```php
class FileDoesNotExistException extends Exception
{
	protected $message = 'File does not exist';
}
```

php/src/vaas/Exceptions/VaasAuthenticationException.php
```diff
@@ -6,4 +6,8 @@
 
 class VaasAuthenticationException extends Exception
 {
-}
+    public function __construct(string $message = 'Authentication failed', int $code = 0, Exception $previous = null)
```
unglaublicherdude, line 9: same here

php/src/vaas/Exceptions/VaasClientException.php
```diff
@@ -4,20 +4,10 @@
...
-/**
- * The server encountered an internal error.
- * Recommended actions:
- * * You may retry the request after a certain delay.
- * * If the problem persists contact G DATA.
- */
 class VaasClientException extends Exception
 {
-    public function __construct(?string $message)
+    public function __construct(string $message = 'Client error', int $code = 0, Exception $previous = null)
```
unglaublicherdude, line 9: same here

php/src/vaas/Options/ForSha256Options.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace VaasSdk\Options;
+
+class ForSha256Options
+{
+    public bool $useCache;
+    public bool $useHashLookup;
+    public ?string $vaasRequestId;
+
+    public function __construct(array $options = [])
```
unglaublicherdude, line 11: Why do you use an arbitrary array here instead of just parameters for the constructor?

php/src/vaas/Options/ForStreamOptions.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace VaasSdk\Options;
+
+class ForStreamOptions
+{
+    public bool $useHashLookup;
+    public int $timeout;
+    public ?string $vaasRequestId;
+
+    public function __construct(array $options = [])
```
unglaublicherdude, line 11: same

php/src/vaas/Options/ForUrlOptions.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace VaasSdk\Options;
+
+class ForUrlOptions
+{
+    public bool $useHashLookup;
+    public ?string $vaasRequestId;
+
+    public function __construct(array $options = [])
```
unglaublicherdude, line 10: same

php/src/vaas/VaasBuilder.php
unglaublicherdude, line 1: nice!

php/tests/VaasTesting/AuthenticatorTest.php
unglaublicherdude, line 1: Why did you move the path?
So the source is now in src/vaas but the test are in tests/VaasTesting. Any reason for that?

lennartdohmann: So its psr conform

unglaublicherdude (approved):
This is already tested with the nextcloud app right? Do we have data of the implication of the performance on this change? And can we actually release it now?

unglaublicherdude:
Ah. And maybe after 3 weeks we should update the dependencies here.

php/src/vaas/Vaas.php
```diff
@@ -2,51 +2,47 @@
...
-        if (isset($this->vaasConnection)) {
-            $this->vaasConnection->close();
-        }
+    public static function builder(): VaasBuilder {
+        return new VaasBuilder();
+    }
+
+    public static function createInstance(): Vaas {
+        return new self();
     }
```
GermanCoding, line 46: I'm a bit unhappy that these functions (+ the with*** functions) are public, because that means that a library user can just bypass the builder completely. To prevent this we would need to restrict visibility here, but then the builder needs to be related to Vaas or it won't be able to use the constructor either. Not sure what's best here.

php/src/vaas/Vaas.php
```diff
@@ -73,314 +69,321 @@ public function withLogger(LoggerInterface $logger): self
...
-        return new VaasVerdict(
-            $this->_verdictResponseForSha256(
+            $url = sprintf('%s/files/%s/report/?useCache=%s&useHashLookup=%s',
+                $this->options->vaasUrl,
                 $sha256,
-                $uuid
-            )
-        );
+                json_encode($options->useCache),
+                json_encode($options->useHashLookup
```
GermanCoding, line 90: While it doesn't matter for booleans, `urlencode()` would be the right thing here, not JSON encoding

php/src/vaas/Vaas.php
```diff
@@ -73,314 +69,321 @@ public function withLogger(LoggerInterface $logger): self
...
-        ));
-    }
+            $options = $options ?? ForFileOptions::fromVaasOptions($this->options);
+
+            if ($options->useCache || $options->useHashLookup) {
+                $forSha256Options = new ForSha256Options(
+                    $options->useCache, $options->useHashLookup, $options->vaasRequestId);
+                $sha256 = Sha256::TryFromFile($path);
+                $this->logger->debug("Check if file $path is already known by its SHA256: $sha256");
+                $response = $this->forSha256Async($sha256, $forSha256Options, $cancellation)->await();
```
GermanCoding, line 149: We should catch any errors here. For instance, there could be a failed analysis for this SHA256 hash, which would be returned as an error from VaaS. However this doesn't matter to this function, as it can just re-submit this file. This function should only error if the upload/analysis of the uploaded file (also) fails, but not if just the initial lookup errors.

php/src/vaas/Vaas.php
```diff
@@ -73,314 +69,321 @@ public function withLogger(LoggerInterface $logger): self
...
-            );
-
-            $fileStream->close();
-
-            return $verdict;
-        }
+            $stream = openFile($path, 'r');
 
-        return new VaasVerdict($verdictResponse);
+            $forStreamOptions = new ForStreamOptions($options->useHashLookup, 300, $options->vaasRequestId);
```
GermanCoding, line 159: Magic number, just use the default?

php/src/vaas/Options/ForStreamOptions.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace VaasSdk\Options;
+
+class ForStreamOptions
+{
+    const DEFAULT_TIMEOUT = 300;
+    const DEFAULT_REQUEST_ID = null;
```
GermanCoding, line 8: Consider restricting the visibility of these, or is it relevant to users that these defaults are known?

php/src/vaas/Vaas.php
```diff
@@ -73,314 +69,321 @@ public function withLogger(LoggerInterface $logger): self
...
+                $this->logger->debug("Requesting verdict for stream");
+
+                if (!$stream->isReadable() || $stream->isClosed()) {
+                    $this->logger->error("Stream is not readable");
+                    throw new VaasClientException('Stream is not readable');
+                }
+
+                $options = $options ?? ForStreamOptions::fromVaasOptions($this->options);
+                
+                $url = sprintf('%s/files?useHashLookup=%s', $this->options->vaasUrl, json_encode($options->useHashLookup));
```
GermanCoding, line 187: Same here

GermanCoding (changes requested):
Review part 1, some things I would like to see fixed before we release this.

php/src/vaas/Vaas.php
```diff
@@ -73,314 +69,321 @@ public function withLogger(LoggerInterface $logger): self
...
+
+                $request = new Request($url, 'POST');
+
+                $request->setBody(StreamedContent::fromStream($stream, $fileSize));
+                $request->setTransferTimeout($options->timeout);
```
GermanCoding, line 192: This timeout along with its 300s default effectively means that users with slow connections can never upload medium to large files. Are we confident that this is the default behavior we want? Is there a significant harm to not time-outing uploads?

GermanCoding (changes requested):
Part 2

php/src/vaas/Vaas.php
unglaublicherdude: I am not sure if I get what you are saying. Do you mind opening a PR on this PR with your suggestion?

php/src/vaas/Authentication/TokenReceiver.php
```diff
@@ -0,0 +1,137 @@
...
+                'client_secret' => $this->authenticator->clientSecret,
+                'grant_type' => 'client_credentials',
+            ]),
+            $this->authenticator instanceof ResourceOwnerPasswordGrantAuthenticator =>  http_build_query([
+                'client_id' => $this->authenticator->clientId,
+                'username' => $this->authenticator->userName,
+                'password' => $this->authenticator->password,
+            'grant_type' => 'password',
+            ]),
+            default => throw new InvalidArgumentException("Unknown authenticator type")
```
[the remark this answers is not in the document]
unglaublicherdude: That was actually one of my considerations when I made a PR on this PR to change the old behaviour. Maybe I create a PR with this suggestion so we can compare it.

unglaublicherdude:
?Am 14.01.2025 11:55 schrieb Lennart Dohmann ***@***.***>:
Assigned #678 to @unglaublicherdude.

php/src/vaas/Vaas.php
lennartdohmann: I think PHP does not offer such a way

php/src/vaas/Vaas.php
lennartdohmann: urlencode() is not compatible with the format our API is expecting

unglaublicherdude:
Is this now mergable?

php/examples/VaasExample/ScanIonosS3Bucket.php
GermanCoding, line 1: Name this file just ScanS3Bucket, as the S3 code isn't actually Ionos specific, right?

GermanCoding (approved):

lennartdohmann:
I would say this SDK is now finished and ready to be merged.

PR #682, opened 2024-12-18
refactors the forXXX-options

unglaublicherdude:


php/src/vaas/Options/ForSha256Options.php
```diff
@@ -4,16 +4,10 @@
...
-    public function __construct(array $options = [])
-    {
-        $this->useCache = $options['useCache'] ?? true;
-        $this->useHashLookup = $options['useHashLookup'] ?? true;
-        $this->vaasRequestId = $options['vaasRequestId'] ?? null;
-    }
+    public function __construct(
+        public bool $useCache = true,
+        public bool $useHashLookup = true,
+        public ?string $vaasRequestId = null) {}
```
lennartdohmann, line 10: Yeah I like this more too, thanks!

php/src/vaas/Vaas.php
```diff
@@ -81,15 +81,8 @@ public function forSha256Async(Sha256 $sha256, ?ForSha256Options $options = null
...
-            if ($options === null) {
-                $options = new ForSha256Options(
-                    [
-                        'vaasRequestId' => null,
-                        'useHashLookup' => $this->options->useHashLookup ?? true,
-                        'useCache' => $this->options->useCache ?? true,
-                    ]
-                );
-            }
+            $options = $options ?? ForSha256Options::default();
```
lennartdohmann, line 84: This will cause a bug. If the local and function specific options are null, we need to introduce new ForSha256Options with the global VaasOptions' settings injected to the local ones, if they are set, exactly like before. Just nullcheck and make default ForSha256Options like you changed it to, would overwrite VaasOptions and could cause VaaS not behave like a customer would like to. I suggest it like it was before but with your updated constructor:
```suggestion
            if ($options === null) {
                $options = new ForSha256Options($this->options->useCache, $this->options->useHashLookup, null);
            }
```

php/src/vaas/Vaas.php
```diff
@@ -146,22 +139,12 @@ public function forFileAsync(string $path, ?ForFileOptions $options = null, ?Can
...
-            if ($options === null) {
-                $options = new ForFileOptions(
-                    [
-                        'vaasRequestId' => null,
-                        'useHashLookup' => $this->options->useHashLookup ?? true,
-                        'useCache' => $this->options->useCache ?? true,
-                    ]
-                );
-            }
+            $options = $options ?? ForFileOptions::default();
```
lennartdohmann, line 142: The same here like in line 84.
```suggestion
            if ($options === null) {
                $options = new ForFileOptions($this->options->useCache, $this->options->useHashLookup, null);
            }
```

php/src/vaas/Vaas.php
```diff
@@ -203,15 +183,8 @@ public function forStreamAsync(ReadableStream $stream, int $fileSize, ?ForStream
...
-                if ($options === null) {
-                    $options = new ForStreamOptions(
-                        [
-                            'vaasRequestId' => null,
-                            'timeout' => $this->options->timeout,
-                            'useHashLookup' => $this->options->useHashLookup ?? true,
-                        ]
-                    );
-                }
+                $options = $options ?? ForStreamOptions::default();
```
lennartdohmann, line 186: Same here as before:
```suggestion
               if ($options === null) {
                    $options = new ForStreamOptions($this->options->useHashLookup, 300, null);
                }
```

php/src/vaas/Vaas.php
```diff
@@ -272,14 +242,8 @@ public function forUrlAsync(string $uri, ?ForUrlOptions $options = null, ?Cancel
...
 
-            if ($options === null) {
-                $options = new ForUrlOptions(
-                    [
-                        'vaasRequestId' => null,
-                        'useHashLookup' => $this->options->useHashLookup ?? true,
-                    ]
-                );
-            }
+            $options = $options ?? ForUrlOptions::default();
```
lennartdohmann, line 245: Same here as before:
```suggestion
            if ($options === null) {
                $options = new ForUrlOptions($this->options->useHashLookup, null);
            }
```

lennartdohmann (changes requested):

unglaublicherdude:
thank you for finding that bug

lennartdohmann (approved):

PR #684, opened 2024-12-25
Refactor cs/http_api

lennartdohmann:
Refactor cs/http_api:
* Same usage as other SDK
* Fix override of VaasOptions if no specific options are passed


dotnet/Vaas/src/Vaas/Authentication/TokenReceiver.cs
```diff
@@ -109,36 +109,36 @@ private async Task<TokenResponse> RequestTokenAsync(CancellationToken cancellati
 
     private FormUrlEncodedContent TokenRequestToForm()
```
GermanCoding, line 110: Consider making this class abstract, and just implement `TokenRequestToForm` in `ClientCredentialsGrantAuthenticator`/`ResourceOwnerPasswordGrantAuthenticator`, thus getting rid of the switch and logically separating the two.

lennartdohmann:
@GermanCoding you likey likey?

GermanCoding (approved):

PR #970, opened 2026-06-01
python/Encrypted files in VaaS

jbg-1:
Implement isEncrypted for Python SDK

### DIFF
--- a/python/tests/test_vaas.py
+++ b/python/tests/test_vaas.py
@@ -24,6 +24,10 @@ PUP_SHA256 = "d6f6c6b9fde37694e12b12009ad11ab9ec8dd0f193e7319c523933bdad8a50ad"
 EICAR_URL = "https://s3-eu-central-2.ionoscloud.com/test-samples-vaas/eicar.com.txt"
 CLEAN_URL = "https://s3-eu-central-2.ionoscloud.com/test-samples-vaas/clean.txt"
 PUP_URL = "https://s3-eu-central-2.ionoscloud.com/test-samples-vaas/PotentiallyUnwanted.exe"
+PASSWORD_ZIP_URL = "https://s3-eu-central-2.ionoscloud.com/test-samples-vaas/password.zip"
+WITH_AND_WITHOUT_PASSWORD_ZIP_URL = (
+    "https://s3-eu-central-2.ionoscloud.com/test-samples-vaas/with-and-without-password.zip"
+)
 
 CLEAN_FILE_CONTENT = "I am clean."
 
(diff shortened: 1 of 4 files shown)


python/tests/test_vaas.py
```diff
@@ -1099,3 +1099,50 @@ async def test_for_url_cancel_request_raise_cancel_error(self, vaas, post_fails,
 
         with pytest.raises(asyncio.CancelledError):
             await vaas.for_url(url)
+
+PASSWORD_ZIP_URL = "https://samples.develop.vaas.gdatasecurity.de/password.zip"
+WITH_AND_WITHOUT_PASSWORD_ZIP_URL = "https://samples.develop.vaas.gdatasecurity.de/with-and-without-password.zip"
+
+
+class TestVaasEncrypted:
```
doxthree: Why own class?

python/tests/test_vaas.py
```diff
@@ -1099,3 +1099,50 @@ async def test_for_url_cancel_request_raise_cancel_error(self, vaas, post_fails,
 
         with pytest.raises(asyncio.CancelledError):
             await vaas.for_url(url)
+
+PASSWORD_ZIP_URL = "https://samples.develop.vaas.gdatasecurity.de/password.zip"
```
doxthree: Better have all url samples on one place like currently on IONOS S3

python/tests/test_vaas.py
```diff
@@ -1099,3 +1099,50 @@ async def test_for_url_cancel_request_raise_cancel_error(self, vaas, post_fails,
...
+            with open(filename, mode="wb") as file:
+                file.write(response.content)
+
+            verdict = await vaas.for_file(filename)
+
+            assert verdict.verdict == "Clean"
+            assert verdict.isEncrypted is True
+
+    @pytest.mark.asyncio()
+    async def test_for_file_if_contains_eicar_and_encrypted_returns_malicious_and_is_encrypted(self, vaas):
```
doxthree: Idea: Add the files, you want to test in the already written test and check for the encryption flag for all samples

doxthree (changes requested):

python/tests/test_vaas.py
GermanCoding: The Java SDK already uses this location (https://github.com/GDATASoftwareAG/vaas/blob/7fe05a010194721038b7a779df7bf04b41b1636a/java/src/test/java/de/gdata/test/integration/RealApiIntegrationTests.java#L54, https://github.com/GDATASoftwareAG/vaas/blob/7fe05a010194721038b7a779df7bf04b41b1636a/java/src/test/java/de/gdata/test/integration/SamplesFixture.java#L29) [due to better uptime I believe]. We should consider moving all to either location.

jbg-1: Samples doesn't provide content-length.
We'll move all urls to s3.

doxthree (approved):

GermanCoding (approved):

PR #971, opened 2026-06-01
rust/Encrypted files in VaaS

jbg-1:
Implement isEncrypted for Rust SDK

### DIFF
--- a/rust/tests/real_api_integration_tests.rs
+++ b/rust/tests/real_api_integration_tests.rs
@@ -1,5 +1,6 @@
 use reqwest::Url;
 use std::io::Write;
+use std::path::Path;
 use std::str::FromStr;
 use std::sync::LazyLock;
 use tokio_util::sync::CancellationToken;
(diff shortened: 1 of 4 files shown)


rust/tests/real_api_integration_tests.rs
```diff
@@ -14,8 +15,26 @@ static EICAR_SHA256: LazyLock<Sha256> = LazyLock::new(|| {
...
+        if candidate.is_file() {
+            dotenv::from_path(Path::new(&candidate)).ok();
+            return;
+        }
+        match dir.parent() {
+            Some(parent) => dir = parent.to_owned(),
+            None => return,
+        }
+    }
+}
```
GermanCoding: ? the dotenv crate already searches parent directories, why is this needed? https://docs.rs/dotenv/latest/dotenv/fn.dotenv.html

rust/tests/real_api_integration_tests.rs
```diff
@@ -14,8 +15,26 @@ static EICAR_SHA256: LazyLock<Sha256> = LazyLock::new(|| {
...
+
+static DOTENV: LazyLock<()> = LazyLock::new(load_dotenv_from_ancestors);
```
GermanCoding: Since we're already using rstest, you can remove this lock (and the static) and convert this into a "once" fixture instead: https://github.com/la10736/rstest#use-once-fixture

rust/tests/real_api_integration_tests.rs
```diff
@@ -25,13 +44,14 @@ fn vaas_with_authenticator(authenticator: impl Authenticator + 'static) -> Vaas
...
 #[rstest::fixture]
 fn vaas_with_client_credentials() -> Vaas {
-    let token_url: Url = dotenv::var("TOKEN_URL")
+    let _ = *DOTENV;
+    let token_url: Url = std::env::var("TOKEN_URL")
         .expect("No TOKEN_URL environment variable set to be used in the integration tests")
         .parse()
         .expect("Failed to parse TOKEN_URL environment variable");
-    let client_id = dotenv::var("CLIENT_ID")
+    let client_id = std::env::var("CLIENT_ID")
```
GermanCoding: the behavior of these two should be equivalent, but dotenv::var is actually safer because it ensures that dotenv was initialized properly

GermanCoding (commented):
What was the issue with the dotenv loading?

rust/tests/real_api_integration_tests.rs
GermanCoding: Actually I just realized that the vaas initializers are already fixtures. You can either nest fixtures with rstest, but I just looked and realized that dotenv already does threadsafe loading so doing this as before should have been fine already.

GermanCoding (approved):

